PR #2, opened 2015-09-18
Correction to allow the LOGO flightplan to start via init,

KeesGuijt:
fix to reduce the sample frequency for analog inputs following the change in HEARTBEAT_HZ (r1172).

### DIFF
--- a/MatrixPilot/flightplan.c
+++ b/MatrixPilot/flightplan.c
@@ -54,14 +54,17 @@ void flightplan_init(void)
 {
 #if (FLIGHT_PLAN_TYPE == FP_LOGO)
 	flightplan_logo_active = true;
-#endif
 	flightplan_logo_init();
+#else	
 	flightplan_waypoints_init();
+#endif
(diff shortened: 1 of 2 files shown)


KeesGuijt:
- r1155 (abstract the flightplan handler) crippled LOGO. Available modes where manual, stabilized, and manual.
  I reported this before: https://groups.google.com/forum/#!topic/uavdevboard/dKxv509p060, apr 16.
  I have not tested FP_WAYPOINTS
- the analog fix adresses that after the change form  HEARTBEAT_HZ=40 to HEARTBEAT_HZ=200 for Udb5 and Auav3,  heartbeat_pulse(void) is run at 200Hz.
  This leaves too little time for analog samples to be completed. The fix brings te sample rate back to 40Hz. I only tested this on a Auav3 board.
  without the fix, the first sample is correct, but no updates follow. This can easily go unnoticed. So testing should be done by changing the applied voltage repeatedly.

PR #3, opened 2015-09-18
reduce the sample frequency for analog inputs following the change in HEARTBEAT_HZ (r1172).

KeesGuijt:
fix to reduce the sample frequency for analog inputs following the change in HEARTBEAT_HZ (r1172).

- r1155 (abstract the flightplan handler) crippled LOGO. Available modes where manual, stabilized, and manual.
  I reported this before: https://groups.google.com/forum/#!topic/uavdevboard/dKxv509p060, apr 16.
  I have not tested FP_WAYPOINTS
- the analog fix adresses that after the change form  HEARTBEAT_HZ=40 to HEARTBEAT_HZ=200 for Udb5 and Auav3,  heartbeat_pulse(void) is run at 200Hz.
  This leaves too little time for analog samples to be completed. The fix brings te sample rate back to 40Hz. I only tested this on a Auav3 board.
  without the fix, the first sample is correct, but no updates follow. This can easily go unnoticed. So testing should be done by changing the applied voltage repeatedly.

### DIFF
--- a/libUDB/heartbeat.c
+++ b/libUDB/heartbeat.c
@@ -117,9 +117,12 @@ static void heartbeat_pulse(void)
 	vref_adj = 0;
 #endif // VREF
 
-	calculate_analog_sensor_values();
 	udb_callback_read_sensors();
-	udb_flags._.a2d_read = 1; // signal the A/D to start the next summation
+	if ((udb_heartbeat_counter % (HEARTBEAT_HZ/40)) == 1)
+	{
+		calculate_analog_sensor_values();
+		udb_flags._.a2d_read = 1; // signal the A/D to start the next summation
+	}
 
 	// process sensor data, run flight controller, generate outputs. implemented in libDCM.c
 	udb_heartbeat_callback(); // this was called udb_servo_callback_prepare_outputs()


phollands:
Bill,

Can you please confirm or comment on  the following:-

I know that you recently attributed the high CPU load (>40%) in trunk, to the use of a 200Hz Heartbeat in trunk only for the AUAV3 and the UDB5 (not the UDB4 which retains a 40HZ Heartbeat). I think you then put in a test of the UDB4 and AUAV3 with 40HZ Heartbeat and the cpu came back down to 24% load. You were of the opinion planes should really stay at 40HZ and that only quad copters and other copters needs the 200HZ frequency.

To summarise, you fix for MatrixPilot for planes, would be to keep the Heartbeat at 40Hz and not use 200Hz. Is that correct ?

Best wishes, Pete

wpremerlani:
Peter,
Yes, I would set the heartbeat at 40 Hz for all UDB types for fixed wing.
I have been doing lots of testing lately with my EasyStar2 with both the
200 Hz heartbeat and 40 Hz heartbeat, and cannot detect any difference in
performance.
That said, there is no downside to the 200 Hz heartbeat, other than heavy
CPU usage.
Best regards,
Bill

KeesGuijt:
There is one downside for 200Hz heartbeat; it leaves too little time for analog samples to be completed.

I wonder what the comment with r1172 means, and if it still is true:
"for legacy purposes UDB4 board still defaults to 40Hz heartbeat whilst all MPU6000 based targets require default (unless MPU6000 driver is extended) of 200Hz heartbeat. Therefore configure this in the code."

alternatively, it could be decided to revert r1172, and go back to 40Hz as default in trunk.

Inspirati:
G'day Kees,

There was a fix developed for the analog sampling rate, quite some time
ago. Unfortunately I don't recall who did the work, or which branch i'd
best search in for it..

The MPU6000 driver was developed with 200Hz hard coded in the setup
routine. It could be changed, however non of us have yet ventured there.
Credit goes to MarkW for the original development work.

Therefore we can't quite so easily simply revert to any particular revision
in order to address this - one will have to dig a little deeper (-;

Best wishes, Robert.

KeesGuijt:
Hi Robert,

Thanks for your comment. The fix I propose leaves the heartbeat at 200hz, while restoring analog sampling to 40Hz. Of course,I meant only to revert one revision, not revert to r1171 :). 
So running MPU6000 on 40Hz works, but is not optimal?

All the best,
Kees

phollands:
Kees, Thanks for updating the title and the commit. Please can you tell me a little about what testing you have done with this fix.

 Telemetry makes sense  with current used etc OK ? Did CPU usage go down much ? I assume you are running at 16 MIPS ? And are you testing on the AUAV3 or UDB5 ?

KeesGuijt:
Hi Pete,

I've been using this fix for over a year now. I use two AUAV3 for flying, and use an analog input with my own telemetry fixes to readout the voltage on the battery. I believe my UDB5 board had the same problem, but I'm not sure. I could test it for you, if you like. I run at 16 MIPS. I have not looked at the impact on the cpu load, but that will probably be small.

Regards,
Kees

phollands:
Kees, I've not forgotten about this. I'm hoping to wire up the correct hardware and test this with the UDB5 in my plane. If anyone else get's to test it first, then that is fine, and I'll commit this. It will probably be another few weeks before I get around to this. Best wishes, Pete

phollands:
Kees, just to let you know I looked at this a little more  today.

From that I can see the exact commit to which you are referring is:-

> svn MatrixPilot_trunk: for legacy purposes UDB4 board still defaults to 40Hz heartbeat whilst all MPU6000 based targets require default (unless MPU6000 driver is extended) of 200Hz heartbeat. Therefore configure this in the code. NOTE: as this setting depends on a C preprocessor 'project wide' define, **potential for a hidden build bug is high**.. (aka, undetected HEARTBEAT_HZ defined to different values in different modules)
> 
> git-svn-id: https://gentlenav.googlecode.com/svn/trunk@3472 cf4e4d4a-e611-11dd-8fb1-9960f2a117f8
> Commit: [ee1814ce8470d910d9e052f1176ade2aa0b2c4c8 [ee1814c]](https://github.com/MatrixPilot/MatrixPilot/commit/ee1814ce8470d910d9e052f1176ade2aa0b2c4c8)
> Parents:  de282d1150
> Author:  Robert Dickenson <robert.dickenson@gmail.com>
> Date: 31 January 2015 06:27:11 GMT

Should the call to  calculate_analog_sensors() also be changed to be at 40Hz ?
Can you add that in ? I think we are then "good to go".

Best wishes, Pete

KeesGuijt:
Hi Pete,

The fix is now like i first commited it, with the three lines to 40Hz.

All the best,
Kees

Ps. Nice to see so much good activities, thanks.

phollands:
Actually, my understanding of this that the call to udb_callback_read_sensors(); needs to be at the full HEARTBEAT_HZ rate.  That call is reading the average gyro and accelerometer readings to use in the next DCM update (Operating at HEARTBEAT_HZ). So those values, for the UDB4 and AUAV3 are coming in from the MPU6000 at 200Hz, and need to be read for the DCM at 200Hz. 

It is a little bit confusing that the the line udb_callback_read_sensors(); is embedded between the two analogue sensor lines, and so the call appears to be associated with the analogue readings, when in fact, for the UDB5 and AUAV3, the call refers to sensors that arrive via a separate digital channel.

So, if you agree that I have understood this correctly, then we could move udb_callback_read_sensors() up a couple of lines (out of the 40HZ loop).

KeesGuijt:
Pete, right u are.

wpremerlani:
Peter,
I assume that you had a typo in your last email. You said:
"So those values, for the UDB4 and AUAV3 are coming in from the MPU6000 at
200Hz"
The UDB4 does not have the MPU6000. I think you meant to say:
"So those values, for the UDB5 and AUAV3 are coming in from the MPU6000 at
200Hz"
Best regards,
Bill

phollands:
Yes, that was typo. sorry about that. I meant UDB5 and AUAV3.

phollands:
Kees, I found that the commit broke the UDB4 xrate.value etc (summation / average values). So I have reverted the commit on master.

KeesGuijt:
Pete.

Too bad. Do you think it is specific to only UDB4? Which of the two line caused that? I'll do some digging too.

All the best,
Kees

phollands:
Yes specific only to the UDB4 with respect to the RollPitchYaw Test. The UDB5 and AUAV5 tested fine. 
I'm confident I'll find the issue tomorrow.

phollands:
Kees, the issues was that in your patch we had:-
`if ((udb_heartbeat_counter % (HEARTBEAT_HZ/40)) == 1)`
instead of
`if ((udb_heartbeat_counter % (HEARTBEAT_HZ/40)) == 0)`
The first statement can never be satisfied when HEARTBEAT_HZ has a value of  40; as anything modulo 1 can never have a result of anything but 0, so it was never going to == 0. The code in the if block never executed.
So it's an easy fix.  I'll push the fix. I  want to test my UDB4 gyros initialisation (auto-zero) branch on top of this fix today.

KeesGuijt:
Pete, 
Great. thanks.

PR #4, opened 2015-09-21
Update with final commits to googlecode SVN repository

Inspirati:
comments as per individual commit messages.


phollands:
Robert, Thanks.

PR #10, opened 2016-04-22
Change default of USE_MCU_IDLE from 1 to 0 .

phollands:
As[ discussed recently in uavdevboard](https://groups.google.com/d/msg/uavdevboard/Ja-NeeXkC50/vHyuhmoLAwAJ) I am moving to have USE_MCU_IDLE be switched off i.e. move from being 1 to 0.

### DIFF
--- a/libUDB/interrupt.h
+++ b/libUDB/interrupt.h
@@ -21,7 +21,7 @@
 
 ////////////////////////////////////////////////////////////////////////////////
 // Optionally enable the new power saving idle mode of the MCU during mainloop
-#define USE_MCU_IDLE    1
+#define USE_MCU_IDLE    0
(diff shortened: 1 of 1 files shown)


phollands:
Flight tested today in my Autopilot stack on an Alpha Axion. Flight was OK and fine. However CPU usage is high on the UDB5 and AUAV3. This was similar to a result from an earlier flight with use USE_MCU_IDLE set to 1, which i [reported on uavdevboard.](https://groups.google.com/d/msg/uavdevboard/xWHRCKIFewQ/M4FHbYnJBAAJ)  I will go ahead and merge this change, and continue the CPU loading discussion elsewhere.

PR #12, opened 2016-04-23
Add battery readings, voltage, ampage and milli amp hours, to csv spr…

phollands:
…eadsheet produced by flan.pyw. This commit tested against telemetry from flight 191 by Giulio Berti. 
battery_readings_in_spreadsheet


phollands:
On uavdevboard, Giulio has written: The CSV is perfect, you even scaled the values!! :DDD
And then generated a graph in the spreadsheet.

So we are "Good to Go"

PR #19, opened 2016-05-26
Include battery Voltage and Current in the Mavlink SYS_STATUS mesage

giukio:
As I started to use the Minim Osd, flashed with MinimOsdExtra firmware, I noticed that voltage and current data were not displayed, even though I had already configured the voltage and current sensor in options.h

I wanted to fix this also to learn the GitHub workflow, explained by Pete in [one of his videos](https://vimeo.com/163824531).

For the time being, I've only compiled this fix, I'll fly test it tomorrow.

Personally, I find that the use of #if () #else #endif disrupts the readability of the code, but I understand the memory efficiency needs.

Best regards,
Giulio

### DIFF
--- a/MatrixPilot/MAVLink.c
+++ b/MatrixPilot/MAVLink.c
@@ -975,7 +975,7 @@ void mavlink_output_40hz(void)
             #else
 				(int16_t)0,                    
             #endif
-		    0,                              // Percentage battery remaining 100 percent is 1000
+		    100,                               // Remaining battery energy: (0%: 0, 100%: 100), -1: autopilot estimate the remaining battery
 		    r_mavlink_status.packet_rx_drop_count,
 		    0,              // errors_comm
 		    0,              // errors_count1
--- a/MatrixPilot/MAVLink.c
+++ b/MatrixPilot/MAVLink.c
@@ -965,9 +965,17 @@ void mavlink_output_40hz(void)
 		    0,              // Sensors enabled
 		    0,              // Sensor health
 		    udb_cpu_load() * 10,
-		    0,              // Battery voltage in mV
-		    0,              // Current
-		    0,              // Percentage battery remaining 100 percent is 1000
+            #if (ANALOG_VOLTAGE_INPUT_CHANNEL != CHANNEL_UNUSED)
+                battery_voltage._.W1 * 100,     // Battery voltage, in millivolts (1 = 1 millivolt)
+            #else
+                (int16_t)0,
+            #endif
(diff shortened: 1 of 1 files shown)


phollands:
Giulio. Thanks very much. If this passes your test fly tomorrow, then we'll merge. Code looks good to me.
Best wishes, Pete

magicrub:
I'd suggest moving the #if to where battery_voltage._.W1 gets populated. That would fix it in all other cases in the codebase that we haven't caught yet

giukio:
Hi Tom, 
if I understood you correctly, the #if are already present where battery_voltage, battery_current and also battery_mAh_used are written. 

By removing the #if in the other parts of the code, we should also remove those around those variable declarations (at the beginning of analogs.c). As far as I understand (probably not very far given that I'm not an experienced programmer) this would have the positive effect of increasing the readability of the code, but the disadvantage of declaring variables even if they aren't actually used, thus occupying memory.

magicrub:
Firstly, it uses an extremely tiny amount of memory so that is not a problem. When we enable that variable it will consume that memory anyway so enabling it all the time would give then added benefit of knowing if that feature is broken by running out of memory.

phollands:
Giulio, you may way want to switch on the diagnostics to see how much of the UDB4's memory you are using for your configuration. That might give you a data point on whether you are short of memory. Here is a screenshot of the location in MPLAB-X for switching on a view of the memory used. You would need to tick "Display memory usage". 

image

After compiling, I get statements like ....
 Total program memory used (bytes):        0x228a5  (141477) 54%

Total data memory used (bytes):         0x1be8  (7144) 23%

Dynamic Memory Usage

region                     address                      maximum length  (dec)
------                     -------                      ---------------------
heap                        0x22a8                               0x100  (256)
stack                       0x23a8                              0x5c18  (23576)

                        Maximum dynamic memory (bytes):         0x5d18  (23832)

The numbers will alter according to the options.h configuration which effects whether MAVLink is included etc. and the variables that we are talking about.

Best wishes, Pete

magicrub:
For the record, this PR is correct with regard to following the president of how it's implemented everywhere else in the codebase. I'm arguing that it's implemented wrong everywhere.

giukio:
Thanks for your comments about memory usage, I get similar values.

The modified SYS_STATUS message showed voltage and current correctly in the OSD, but the remaining battery value had to be changed from 0 to 100 otherwise MinimOsdExtra shows a low battery warning.

Best regards,
Giulio

PR #21, opened 2016-05-28
Navigation light support

giukio:
Hi everyone,

in this branch I added an initial support for led navigation lights in MatrixPilot.
I made a [video introducing the functionalities](https://youtu.be/p2I5vmy2RJ0) (please excuse the language mistakes, I was nervous while speaking english)
Essentially I generalized how leds are switched on and off. 
instead of directly switch on, off, or toggle an led using functions like udb_led_toggle(LED_RED), each led can be dynamically configured by calling the led_wing_set_mode(led,mode,param) function, and it is switched accordingly to its selected mode inside the function led_wing_heartbeat_callback(), called at 20Hz from the low priority heartbeat.

**led_wing_set_mode(led,mode,param)**

- **led**: first associate the led number with name in [led_wing.h](https://github.com/giukio/MatrixPilot/blob/led_wing/libUDB/led_wing.h#L34).
Optionally you can speify LED_ALL_WING to apply the mode to all defined leds (I find this useful).

- **param**: parameter needed by certain led modes

- **mode**
The led modes I defined are:
LED_MODE_OFF: the selected led will be switched off
LED_MODE_ON: the selected led will be switched on
LED_MODE_BLINK: the selected led will blink at frequency specified in param (range clipped at 1-10Hz)
LED_MODE_PATTERN: led normally off, 4 rapid flashes every second. the param fiels is used as offset:as example by setting it to 8 you can make the right wing to flash right after the left one. If you live it at 0 all led in this mode will be synchonized, even if their mode is set at different times.
LED_MODE_PATTERN_INV: same as above, but led normally on, 4 rapid flashes every second

Personally I don't consider led_wing_heartbeat_callback() as finished: I would like to serve in a for loop all the defined leds, but I don't know how to implement a pointer array of output pins.

These modes are then used in states.c

- After the battery is plugged in, before the radio is switched on, all leds are on (to check whether they are working)

- During calibration they blink at 1Hz. when calibration is done, the left wing (red led) will stay on (visual feedback)

- after the 4 aileron waggles, also the right wing (green led) will stay on

- In manual mode both led are on solid

- stabilize uses the inverted pattern: is still a semi-manual mode, thus the led are normally on (like manual), but they rapidly flash 4 times each second

- LOGO or waypoint mode use the normal pattern: led normally off, 4 rapid flashes each second.

- Finally if the plane looses the radio link, all leds will blink at 4Hz. The idea was to give a sort of visual "alarm".

Note that this led mode is [enabled in options.h](https://github.com/giukio/MatrixPilot/blob/led_wing/Config/options.h#L764), and the led_wing_set_mode can be called without #if( USE_WING_LED == 1)

I am open to comments and suggestions, both for the implementation and the choice of the names.
Do you think this could be useful to anyone?

Best regards,
Giulio

### DIFF
--- a/Config/options.h
+++ b/Config/options.h
@@ -760,6 +760,8 @@
 // Define USE_DEBUG_IO to enable DPRINT macro to call printf(..)
 //#define USE_DEBUG_IO
 
+// Enable support for external wing led, currently supported only by UDB4 and UDB5
+#define USE_WING_LED 1
 
 ////////////////////////////////////////////////////////////////////////////////
 // AUAV3 only options
--- a/MatrixPilot/states.c
+++ b/MatrixPilot/states.c
@@ -179,6 +179,7 @@ static void ent_calibrateS(void)
 	stateS = &calibrateS;
 	calib_timer = CALIB_PAUSE;
 	led_on(LED_RED); // turn on mode led
+    led_wing_set_mode(LED_ALL_WING, LED_MODE_ON, 0);
 }
 
 // Acquire state is used to wait for the GPS to achieve lock.
@@ -209,6 +210,7 @@ static void ent_acquiringS(void)
 	stateS = &acquiringS;
 	standby_timer = STANDBY_PAUSE;
 	led_off(LED_RED);
+    led_wing_set_mode(LED_LEFT_WING, LED_MODE_ON, 0);
 }
 
 //	Manual state is used for direct pass-through control from radio to servos.
@@ -223,6 +225,7 @@ static void ent_manualS(void)
 	state_flags._.disable_throttle = 0;
 	waggle = 0;
 	led_off(LED_RED);
+    led_wing_set_mode(LED_ALL_WING, LED_MODE_ON, 0);
 	stateS = &manualS;
 }
 
@@ -244,6 +247,7 @@ static void ent_stabilizedS(void)
 	state_flags._.altitude_hold_pitch = (settings._.AltitudeholdStabilized == AH_FULL || settings._.AltitudeholdStabilized == AH_PITCH_ONLY);
 	waggle = 0;
 	led_on(LED_RED);
+    led_wing_set_mode(LED_ALL_WING, LED_MODE_PATTERN_INV, 0);
 	stateS = &stabilizedS;
 }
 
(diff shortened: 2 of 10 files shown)


giukio:
I forgot to mention that I chose output pins A5 and A14 as the seemed free to me.

they drive an [uln2003a board like this](http://www.dx.com/p/apm-2-5-2-5-2-apm-2-6-diy-led-light-switch-control-module-apm-indicator-module-green-341190#.V0olffmLTIU)

phollands:
Giulio,

I think you could publish the text of your message (cut and paste) to the
uavdevboard discussion group. The only people who will directly be notified
of your pull request are currently myself and Rob, and then I think Kees
and Tom probably have a "Watch" on the pull requests so will also be
notified. (I'm not sure if Bill is getting Pull Request messages yet).  I
think that nearly everyone would love to see your video demonstration on
uavdevboard. Or if you just want developers, then uavdevboard-dev

I note that the automated Travis build for this patch is failing for
Software in the Loop (SIL) and for the Auav3. But as the pull request is
currently just a means for soliciting feature / design comments, that is
fine. Thanks Rob for the Travis build system. That's a great example of it
working well.

Best wishes, Pete

giukio:
Hi Pete,
thanks for your comment :)

The build for Auav3 fails because LED_GREEN_EXT are defined only in configUDB4.h and 5.h
+#define LED_GREEN_EXT       _LATA5
 +#define LED_RED_EXT         _LATA14

Travis did a really good job! thanks Robert!

Best regards,
Giulio

KeesGuijt:
Hi Giulio,

I'm ok with the lighting option, eventhough i don't intend to us it myself. 

I have one comment to make: As an extra, the option should be off by default.

#define USE_WING_LED 0

Btw, I wonder if this would fool Travis?

Best Regards,
Kees

phollands:
Giullo

Although i appreciate the coding that went into this request, I don't see much comment or support for this feature. 

I personally find it hard to distinguish the various modes[ when looking at your video. ](https://youtu.be/p2I5vmy2RJ0?t=128). 

The Travis build is also broken.

 This request has now had little activity since June 1st.  So, unless you would like to proceed further with this, I would like to close this request soon. Do you agree ?

I am appreciative of all of your contributions to the project earlier this year (Testing, Videos and Dashware). Thanks so much.

giukio:
Hi Peter,
I understand that it's not the most useful feature or at least a common
one. I'll continue to use it myself but probably isn't worth the hassle of
implementing it.

Best regards,
Giulio

Il 17 set 2016 10:24, "Pete Hollands" <notifications@github.com> ha scritto:

PR #24, opened 2016-06-08
Update logo and waypoint editor tools to fix broken links (images)

excalbian:
- Updated the URLs for markers to use the GitHub-hosted copy from
googlemaps rather than the (deprecated) GoogleCode URLs.

- Changed the reference to the plane image (plane.png) to use the local
file, which prevents having to update when the project hosting changes.

- Fixed the reference link to point to the GitHub-hosted MatrixPilot

### DIFF
--- a/Tools/UDB_Logo_Editor/udb_logo_editor.html
+++ b/Tools/UDB_Logo_Editor/udb_logo_editor.html
@@ -132,7 +132,7 @@
 	var oldPlanePointLng;
 	
     function setPlaneMarker(){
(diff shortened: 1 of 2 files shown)


phollands:
Adam, Thanks very much for that !. Will just do some testing of my own. Pete

phollands:
All looking good in Chrome  Version 50.0.2661.102 (64-bit) on Imac. Tested Logo Editor and Waypoints Editor.
screen shot 2016-06-08 at 12 02 41

PR #27, opened 2016-07-18
Fix use of MINIMUM_AIRSPEED in two different contexts.

phollands:
MINIMUM_AIRSPEED was used in airspeed_options.h and in helicalTurnCntrl.c in two different contexts, with two different values. 

The compiler was warning that MINIMUM_AIRSPEED was being re-defined. 

This is a small fix for that.


phollands:
Bill has approved this Request.

PR #29, opened 2016-08-03
Fixes to improve compilation of FlashOSD tool

phollands:
This fix does not include the change required to the configuration of the pre-processor options in MLAB or MPLAB-x which is to add an include directory of "../../../Config" (Parent ..s are three levels up).

But with that addition, and these changes the code compiles but has a warning ...
../../../libUDB/heartbeat.c: In function 'heartbeat_pulse':
../../../libUDB/heartbeat.c:121:2: warning: implicit declaration of function 'udb_callback_read_sensors'

Ideally, a developer with working knowledge of the OSD toolkit, and working hardware, would check this out for Ric. who setting up his plane of FPV work.

Best wishes, Pete

### DIFF
--- a/Tools/FlashOSD/main.c
+++ b/Tools/FlashOSD/main.c
@@ -22,6 +22,7 @@
 #include "../../libUDB/libUDB.h"
 #include "../../libUDB/heartbeat.h"
 #include "../../libUDB/osd.h"
+#include "../../libUDB/ADchannel.h"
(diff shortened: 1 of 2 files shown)


Inspirati:
Concur with fixes. These are the same changes which I made in order to restore the build.

I believe we should update the MPLAB-X project as well.

PR #31, opened 2016-08-09
Fix libDCM logic for calibration of sensors; update RollPitchYaw project 

phollands:
Discussion and Experiments on how best to fix up RollPitchYaw in master.

The gyros and accel sensors are not having their offsets stored in the standard version of master. That breaks the demo / test.


libDCM/libDCM.c
```diff
@@ -180,10 +180,11 @@ void udb_heartbeat_callback(void)
 void dcm_calibrate(void)
 {
 	// Don't allow re/calibrating before the initial calibration period has finished
-	if (dcm_flags._.calib_finished)
-	{
+	// This FIX ONLY Experimental for ROllPitchYaw, until the calling of dcm_calibrate is fully understood.
```
magicrub: this should be done in a separate commit which can later be reverted. Do not commit comment code

phollands: Thanks for the Tip.

phollands:
I investigated the settings for White Space in Netbeans / MPLAB-X.  It was defaulting to using a style of tabs / indents called Netbeans which turns Tabs into 4 spaces. I have selected a style of Linux which seems to be the closest to MatrixPilot.

screen shot 2016-08-11 at 07 49 37

phollands:
Tested calibration with Radio and without Radio. All working.

PR #11, opened 2016-04-23
Uncomment the MAVLINK_BAUD macro and assign 57600, as the MAVLink def…

KeesGuijt:
…ault serial speed, Setup MAVLink initialisation. This change may be incomplete.

PR #32, opened 2016-08-11
Extra aircraft type for aircraft with flaps and brakes.

KeesGuijt:
Uses servoMix.ini for configuration. Supports adverse yaw, reduced roll on flaps, butterfly braking and switching between throttle and brakes on one stick using a switch or slider on a different channel. Overspeed protection; braking starts above 150% of desiredSpeed (needs ALTITUDE_GAINS_VARIABLE 1)


phollands:
Kees, for the record, please can you elaborate on the amount of testing and flying that you have done with this code.

KeesGuijt:
Hi Pete,

As you know i've flow many flight with most of this code. The last weeks the code has changed somewhat to support other types of aircraft configurations. I tested the following with Hilsim:
- all features enabled (my normal config)
- the same config but with separated brake and trottle inputs
- glider with ailerons, no motor, no flaps
i've tested those with and without overspeed brakes for 5-10 minutes per combination.

Kees.

phollands:
Kees, I've compiled up your branch Glider_airframe_Glider and have the following comments  to discuss:-

- Should ServoMix.h be moved to within the Config Directory now ?
- Can we leave  change:- GAINS_VARIABLE  and ALTITUDE_GAINS_VARIABLE  set to 0 in this commit ?
- If we do that, possibly we could add a warning if the AIRFRAME is AIRFRAME_GLIDER and the above two parameters are not set ? (your choice).
- I am getting the following warning:-

`../../MatrixPilot/servoMix.c:194:17: warning: unused variable 'flapsSelectedTarget'`

 [The flight video is really nice .](https://www.youtube.com/watch?v=8GmXTy03xQY)

KeesGuijt:
Pete,

•Should ServoMix.h be moved to within the Config Directory now ?
Yes, i like that idea, but that would take changes to the project build tool, how does that work?

•Can we leave change:- GAINS_VARIABLE and ALTITUDE_GAINS_VARIABLE set to 0 in this commit ?
 yes, the option is not required, only recommended
 i unintentionally left the modified gain_variables in the pull request

•If we do that, possibly we could add a warning if the AIRFRAME is AIRFRAME_GLIDER and the above two parameters are not set ? (your choice).
 yes, let's add that warning in config_tests.c

•I am getting the following warning:-
 ../../MatrixPilot/servoMix.c:194:17: warning: unused variable 'flapsSelectedTarget'
 - we could do this:
#if (FLAPS_INPUT_CHANNEL != 0 )
	static int16_t flapsSelectedTarget=0;   //resulting flap selection after checking switch/slider and flight modes, normal speed = 0, slow = -1000, high speed = 1000
#endif

Kees

KeesGuijt:
Pete,

About moving the ini-file:
If we move config.ini, an error may occur when compiling with one of the current project files in trunk (missing file)
We discussed this before: [Maintaining project files](https://github.com/MatrixPilot/MatrixPilot/pull/11#issuecomment-215060851)
- we could instruct MP users to regenerate the project files with the build tool as the standard method
- we could generate new projectfiles for trunk, and keep doing that for every new filechange done in trunk.
I think the first, i'm not shure.

Kees

phollands:
Hi Kess, 

I'm back from short trip away from the Internet.

I was aware that we were talking about moving the configuration of the mixing in servomix.h up to the Config directory. But I did not really have config-ini in my sights (Experimental). Are you using that now for your flights ?

I have recently spent time getting the project files for MPLAB-X in shape. In Master, MatrixPIlot, RollPitchYaw and LEDTest should now all compile "out of the box" (default configuration) with a totally clean compile, and no warnings at all when using MPLAB-X with XC16 version 1.26 or higher.  I did not use the project generation scripts for that.  I'll take a look at the scripts over the next few days, to see if there are easy fixes.

Ideally we would generate the project files using the script, and then update the git repo with the new project files, generated by the script. That way, most users will never need know about the project generation script, unless they are doing substantial development work.

Pete

Inspirati:
"Ideally we would generate the project files using the script, and then
update the git repo with the new project files, generated by the script."

This was my design intent with the scripts. I see a major drawback to not
providing pre-created project files being that the script requires python,
and other than for flan, the typical user does not otherwise need to have
python installed to enjoy MatrixPilot.

Inspirati:
I might also add that I believe servoMix.h belongs in the source directory
parallel with servoMix.c

For any configuration settings, a separate file, for example 'servoMix-
config.h', be created, which would go in the ./Config directory.

I'd also like to see ./Config/options.h broken down into several smaller
functional units. There's a few justifications for why, which I'm happy to
argue for if need be. A good starting example would be all the default
variable values (aka, gains etc), to be in a separate configuration
include. These are all later represented in config.ini, so they are not
build options, but rather more-so dynamic runtime configuration settings.

KeesGuijt:
Hi Pete, Robert,

Taking some time off from computers and internet can be very healthy at times.

I mixed (..) up some files there, sorry about that!
Yes, I meant servoMix.h to be moved, But, i agree, let's put the mixer settings in /Config/servoMix_
config.h (let's use an underscore) instead. Pete, do you agree?

About the options.h and config.ini files, good idea. But i think the first priority would be to speed-up the reading, or read the config.ini in a background process first, before we move on.  

Kees.

phollands:
Kess, Yes I agree. Let's put the configuration of the mixing in servoMix_options.h in the Config directory.

phollands:
Kees, I am sorry; you have done so much work to create updates to the project files for MatrixPilot across 6 different scenarios (MPLAB-X and MPLAB, and then the three board). But I we now have a **conflict**. Is it because your branch Glider_airfameGlider needed rebasing first on Master, before your changes ... because I have also updated the configuration files over the last few weeks  and now you have also changed the same files ? Do you agree ?

I cannot see anyway through this, other than to "reset" your commit in Glider_airframeGlider, then rebase the branch on master, and then create all the changes again, and then make the commit, and push (with the force option) back up to your repository.

The xml configuration files, and .mcp files are not really practical to hand edit, or even to patch onto another changed configuration file.

It's a real pain. I know just how much work you will have done to get those projects sorted.

KeesGuijt:
Pete,

No worries, i'll revert, rebase and do them again by hand.

Kees

PR #34, opened 2016-08-15
Record sensor offsets and typical values in SERIAL_UDB_EXTRA telemetry

phollands:
UDB4 Z gyros can change significantly for some bootups. 
We are now using CUSTOM_OFFSETS and need to reasonably sure that they are close to the seen sensor values. 

It seems sensible to monitor this in the telemetry.  The telemetry will capture the actual offsets (whether they are CUSTOM_OFFSETS or not) and print out the instantaneous value of the sensors (allowing for gravity on Z accelerometer) at the time of the telemetry being output (which starts just after calibration).

Here is some example telemetry:-

F21:Sensors=903,113,-1812,-5758,-6223,-6408
F20:Offsets=832,22,-1759,-5762,-6222,-6412
F15:IDA=Alpha Axion 139 3X MP:IDB=AA-PDH-UK-UDB-STACK:
F16:IDC=Pete Hollands:IDD=http://www.diydrones.com/profile/PeterHollands:

The sensor outputs are in the order:
X accel, Y accel, Z accel, X gyro, Y gyro, Z gyro

### DIFF
--- a/MatrixPilot/telemetry.c
+++ b/MatrixPilot/telemetry.c
@@ -40,6 +40,7 @@
 #include "osd_config.h"
 #if (SILSIM != 1)
 #include "../libUDB/libUDB.h" // Needed for access to RCON
+#include "../libUDB/ADchannel.h"
 #endif
 #include "../libUDB/mcu.h"
 //#include "../libDCM/libDCM_internal.h" // Needed for access to internal DCM values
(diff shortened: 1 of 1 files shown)


phollands:
Test Flown on the UDB4, UDB5 and AUAV3. All OK.

PR #35, opened 2016-08-16
Improve initialisation of UDB4 Gyros 

phollands:
UDB4 Z Gyros have been observed having different zero values between reboots which invalidated the use of CUSTOM_OFFSETS for gyros.  

Remove use of CUSTOM_OFFSETS for gyros.
Introduce use of Auto-Zero function for the UDB4 gyros.
Call Auto-Zero once always at least 200 milliseconds after startup.
Recall Auto-Zero just before calibration of MatrixPilot offsets (just before first waggles of ailerons).
Remove compile messages regarding DMA internals (SELECTED_VALUE).

### DIFF
--- a/libUDB/ADchannel.c
+++ b/libUDB/ADchannel.c
@@ -27,53 +27,47 @@ struct ADchannel udb_xaccel, udb_yaccel, udb_zaccel; // x, y, and z acceleromete
 struct ADchannel udb_xrate,  udb_yrate,  udb_zrate;  // x, y, and z gyro channels
 struct ADchannel udb_vref; // reference voltage (deprecated, here for MAVLink compatibility)
 
+#ifdef UDB4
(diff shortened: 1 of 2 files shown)


magicrub:
This is a good example of why breaking up a commit into smaller parts is important. Smaller chunks are easier to revert specific things. Ideally, each of your 5 changes should have been a separate commit. To do that after the fact, you can reset-MIXED and then just recommit your changes, staging only the lines you want for each given change.

phollands:
This code is currently broken and should not be used.  It is failing on some RollPitchYaw tests.  A messier version of the code at this l[ocation / commit](https://github.com/phollands/MatrixPilot/commit/eda1bfa2e03c2f8b0cc62460904cb9d8a72e2833) level, works fine and flew earlier today.

PR #36, opened 2016-08-16
Auto Zero Gyros on the UDB4 and remove use of CUSTOM_OFFSETS for Gyros.

phollands:
In some testing of the UDB4 with Master, it became apparent that in normal usage, sometimes the Z Gyro would startup with substantially different UDB unit calibration values for zero degrees / second. If the board was configured with CUSTOM_OFFSETS (recommended now for master / helicalTurns) this change of calibration could cause the DCM to go Dizzy both before flying in a plane, and in the RollPitchYaw Demo.

In this Pull Request I am introducing the use of the Auto-Zero function in the UDB4's gyros. Tests with Ric's gyros showed that sometime his gyros were calibrating at -19000 UDB units. The maximum is of the order of -32000 for a gyro input (0 volts). Using the Auto-Zero function has meant that his gyros are now calibrating in the normal -6000 UDB range. I enclose a spreadsheet with further analysis.

I welcome plenty of discussion on this Pull Request as it is changing code at the heart of the IMU.

Best wishes, Pete
P.S. Sorry, but github does not accept spreadsheets. I enclose a PDF version. Email me if you want the actual spreadsheet.

[UDB4_Gyro_Analysis.pdf](https://github.com/MatrixPilot/MatrixPilot/files/421389/UDB4_Gyro_Analysis.pdf)

### DIFF
--- a/libUDB/analog2digital_udb4.c
+++ b/libUDB/analog2digital_udb4.c
@@ -111,8 +111,8 @@ uint16_t maxstack = 0;
 
 
 #define GYRO_POWER_UP_TIME		( 300) // No. of Milliseconds to wait for gyros to power up	
-#define AUTO_ZERO_LATCH_TIME		(  10) // No. of executable instructions to wait 4 microseconds
-#define AUTO_ZERO_SETTLE_TIME		(  20) // No. of executable instructions to wait 10 Milliseconds
+#define AUTO_ZERO_LATCH_TIME		(  10) // No. Microseconds to ensure Auto_Zero latch enabled
+#define AUTO_ZERO_SETTLE_TIME		(  20) // No. Milliseconds to ensure Gyros have settled after auto-zero
 
 
 void udb_init_gyros(void)
@@ -122,13 +122,13 @@ void udb_init_gyros(void)
 	_TRISB14 = 0; //  B14 pin made into an output
 	_LATC4 =   0; // Turn off auto-zeroing
 	_LATB14 =  0; // Turn off auto-zeroing
-	delay_ms(GYRO_POWER_UP_TIME);
+	delay_ms(GYRO_POWER_UP_TIME);		// Gyros specs say allow 200 milliseconds for powerup
 	_LATC4 =   1; // Turn on auto-zeroing
 	_LATB14 =  1; // Turn on auto-zeroing
 	delay_us(AUTO_ZERO_LATCH_TIME);         // z gyro spec says wait at least 2 microseconds
 	_LATC4 =   0; // Turn off auto-zeroing
 	_LATB14 =  0; // Turn off auto-zeroing
-	delay_ms(AUTO_ZERO_SETTLE_TIME);        // z gyro spec says wait at least 7 microseconds
+	delay_ms(AUTO_ZERO_SETTLE_TIME);        // z gyro spec says wait at least 7 milliseconds
 }
 
 void udb_init_accelerometer(void)
--- a/libUDB/analog2digital_udb4.c
+++ b/libUDB/analog2digital_udb4.c
@@ -120,8 +120,8 @@ void udb_init_gyros(void)
 	int i = 0;
 	int j = 0;
 	// turn off auto zeroing 
-	_TRISC4  = 0;
-	_TRISB14 = 0;
+	_TRISC4  = 0; //  C4 pin made into an output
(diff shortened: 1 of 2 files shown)


libUDB/analog2digital_udb4.c
```diff
@@ -110,11 +110,31 @@ uint16_t maxstack = 0;
...
+	_LATC4 =   0; // Turn off auto-zeroing
+	_LATB14 =  0; // Turn off auto-zeroing
+	for (j = 0; j < 1000 * MIPS; j++)
+	{
+		for (i = 0; i < GYRO_POWER_UP_TIME; i++);
+	}
+	_LATC4 =   1; // Turn on auto-zeroing
+	_LATB14 =  1; // Turn on auto-zeroing
+	
+	for (i = 0; i < AUTO_ZERO_LATCH_TIME; i++ ); // z gyro spec says wait at least 2 microseconds
```
magicrub: does this get optimized out?

phollands: Well I change the "for loop" to be  macro "delay_us(x)" [delay x microseconds]. If that is what you mean by optimized out, then yes that commit is further down the list of 5 commits.

phollands:
Bill Premerlani made the following comment by email directly to me:-

"By the way, I looked at your pull request on github, I really cannot tell what the new code will be for sure, but I suspect that one aspect of it is that you are turning off the use of prerecorded gyro offsets for all boards. In my opinion, we should do that only for the UDB4, that for the UDB5 we should use prerecorded for everything."

"Also, I would like to take a look to see where the initialization is called, since it will delay the completion of that code for the width of the pulse, which I think is around 1 millisecond. That might (or might not) be long enough to once in a while cause a stack overflow reboot, it all depends on what other routines it blocks. My preference for this sort of thing is to use an interrupt. I realize that in this case that would not be practical."

phollands:
So I have added a commit that ensure that UDB5 and AUAV3 use CUSTOM_OFFSETS with both the Gyros and Accels; while the UDB4 only uses CUSTOM_OFFSETS with the Accels and not the Gyros.

I tried to steer a middle line with all the #ifdef macros in the commit. I simplified the old code; but then duplicated most of the code for the non UDB4 boards because that makes the source easier to read, while still being maintainable. Otherwise the proliferation of pre-processor macros becomes confusing.

phollands:
With regard to the UDB4 gyro initialisation and the use of delays. There are substantial delays in the current commit on branch UDB4_init_gyros4.  The main delay is for [300 milliseconds](https://github.com/phollands/MatrixPilot/blob/UDB4_gyro_init2/libUDB/analog2digital_udb4.c#L113), to make sure that the maximum 200 millisecond specification for Power Up of the Gyros has been achieved. After initialisation there is a 20 millisecond settling time for the gyros (the spec say allow a maximum of 7 milliseconds).

Looking at RollPitchYaw and MatrixPilot, the initialisation code is run from main(), and so any interrupts will take precedence. But in fact, I don't think any interrupts are running at that time in the initialisation either for RollPitchYaw or MatrixPilot.

The main impact of the gyro power up delay in the UDBe specific code,  will be that the PWM inputs and outputs will take nearly a 1/3 of a second on the UDB4 to startup after a reboot. Some may think that too long. On the other hand, zeroing the gyros before then is not possible. 

This zeroing procedure only improves matters if the UDB4 board is kept still immediately after power up. If the board is moved at powerup, the second calibration of offsets of the gyros, done after about 10 seconds, will counter any errors that were introduced. But it is not ideal. Ideally the gyros would be auto-zeroed, immediately before final calibration of offsets (10 seconds after bootup), and just before the waggle of the aileron servos.  Doing that, with interrupts all running by then, is hard. I'm not sure how one would approach that.

The RollPitchYaw test is working .
I have not yet flight tested this code in the AutoPilot Stack.

phollands:
I have coded another branch, in which I auto-zero the gyros successfully between 8 and 9 seconds after bootup, just before the calibration of offsets, and the 1st waggle of the servos at 10 seconds.
The branch is called background_giro_init and the [1st commit is at this location.](https://github.com/phollands/MatrixPilot/commits/background_giro_init). 

I think that branch is probably superior to the branch for this Pull Request. If I here nothing, open a new pull request for the new scheme. (and close this one).

PR #37, opened 2016-08-18
Background UDB Gyro Initialisation

phollands:
Initialise UDB4 Gyros with Auto_Zero feature, just before the DCM offset calibration, about 9 seconds after bootup. There is no use delay functions in this code.

CUSTOM_OFFSETS are not used for the Gyros of the UDB4. Testing had shown that UDB4 gyros could power up with different offset values. Other boards (e.g. UDB5 and AUAV3) continue to use CUSTOM_OFFSETS on their gyros.

This version of the code uses auto-zero timing relative to CALIB_COUNT (now called DCM_CALIB_COUNT) in the dcm routine for recording offsets.

### DIFF
--- a/libDCM/libDCM.c
+++ b/libDCM/libDCM.c
@@ -35,10 +35,6 @@
 union dcm_fbts_word dcm_flags;
 int16_t angleOfAttack;
 
-// Calibrate for 10 seconds before moving servos
-#define CALIB_COUNT  400    // 10 seconds at 40 Hz
-#define GPS_COUNT    1000   // 25 seconds at 40 Hz
-
 void send_HILSIM_outputs(void);
 
 void SetAofA(int16_t AofA)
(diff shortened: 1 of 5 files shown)


phollands:
I would be grateful for any further comments from developer as I have now started a series of flight tests for this branch, and will be loathe to make changes soon.

I have tested this with RollPitchYaw (MPLAB-X IDE, XC16 compiler on Mac) and this passed the normal tests with the three servos, and passed examination of the Sensor Logs. (About 4 tests with each board).

I have test flown this branch with MatrixPilot composed of branch background_giro_init, which in turn was re-based on the latest master.  The UDB4 flew the plane, and looks fine on analysis of the logs. The UDB5 followed through and closely follows the track and orientation of the UDB4. (For this flight the AUAV3 GPS wiring seems not to have worked properly. No GPS data reached the AUAV3.. ) I will re-test with the AUAV3 tomorrow.

wpremerlani:
Peter,
Looks good to me. In any case, if I missed something, I have no one to
blame but myself.
In view of the fact that tests went ok, I support your decision to make
these changes.
Best regards,
Bill

PR #38, opened 2016-08-23
RollPitchYaw: Improve IDE project configuration for MPLAB-X

phollands:
The project was failing to load correctly in MPLAB-X because  FSconfig.h  was configured as being in the libFlashFS directory, but has been moved to the RollPitchYaw/config directory This was effecting some of the IDE parsing and colour highlighting of some files in the editor. This commit fixes that issue.

The commit also deletes an extraneous options.h in the RollPitchYaw directory as there is already an options.h ,one directory lower, in the Config directory.


phollands:
For reference, I enclose a screenshot of how the project looks now.
rollpitchyaw_mplab-x_project_screenshot

PR #39, opened 2016-08-23
AUAV3: Stop Null char  at start of telemetry and GPS UART transmit streams.

phollands:
This commit ensures that a Null Character is not present at the start of every OpenLog File when using the AUAV3.

The original code for the AUAV3, setup the Input and Output pins of the UART as part of the configuration subroutine, configureDigitalIO(void). However,in the reference manual, and UART chapter for this processor, it explains that the pins will be setup when the UART peripheral module is enabled. That will happen automatically at that time.

So the old code setup the pins twice. Once in configureDigitalIO and once when the UART module was enabled. The result was that a Null Character (ascii value \000) was sent down the TX stream.


phollands:
Thanks Tom. I improved the commit as per your suggestion.

PR #16, opened 2016-05-07
Glide airspeed control fix

KeesGuijt:
if one or more of these setting are used, compilation errors may occur, and/or code will not function;

GAINS_VARIABLE                      1
ALTITUDE_GAINS_VARIABLE             1
GLIDE_AIRSPEED_CONTROL  1

fixing this may help to fix SPEED_CONTROL as well.

Best regards,
Kees

PR #40, opened 2016-08-23
AUAV3: A Microchip recommended workaround for Silicon Errata #16

phollands:
The errata in the silicon means that the TX interrupt, which is meant to signify that all characters in the FIFO have been sent down the serial stream, may trigger too soon.  This is the recommended fix.
See the following document for further details:
http://ww1.microchip.com/downloads/en/DeviceDoc/80000526F.pdf
The document says:-

**16. Module: UART**
When using UTXISEL = 01 (interrupt when last
character is shifted out of the Transmit Shift
Register), and the final character is being shifted
out through the Transmit Shift Register (TSR), the
TX interrupt may occur before the final bit is shifted
out.
**Work around**
If it is critical that the interrupt processing occurs
only when all transmit operations are complete,
after which the following work around can be
implemented:
Hold off the interrupt routine processing by adding
a loop at the beginning of the routine that polls the
transmit shift register empty bit, as shown in
Example 2.

### DIFF
--- a/libUDB/serialIO.c
+++ b/libUDB/serialIO.c
@@ -122,6 +122,10 @@ void __attribute__((__interrupt__,__no_auto_psv__)) _U1TXInterrupt(void)
 #endif // HILSIM_USB
 	if (txchar != -1)
 	{
+#if (BOARD_TYPE == AUAV3_BOARD)
+		// a recommended workaround for Silicon Errata #16 in dsPIC33EP512MU810
+		while(U2STAbits.TRMT==0); // wait for the transmit buffer to be empty
+#endif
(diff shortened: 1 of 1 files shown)


phollands:
On Re-reading this issue carefully, I now understand that it will not effect our use of the UART. The issues is saying that the TX interrupt may occur before the final bit has left the serial stream encoder. This does not normally matter to our code. It only matters if the code needs to be absolutely sure that the bit stream generation has been completed. But in our case we are simply wanting to add the next byte to the FIFO registers.

So I am not going to implement this workaround. We may as well keep the code simple.

PR #41, opened 2016-08-23
AUAV3: Prevent issue 17 of silicon Errata 

phollands:
Ensure the STA register BRK has cleared completely before changing any other bits in the STA register.
Further details at http://ww1.microchip.com/downloads/en/DeviceDoc/80000526F.pdf

### DIFF
--- a/libUDB/serialIO.c
+++ b/libUDB/serialIO.c
@@ -66,6 +66,11 @@ void udb_init_GPS(int16_callback_fptr_t tx_fptr, callback_uint8_fptr_t rx_fptr)
 	U1STAbits.UTXISEL0 = 1;     //Bit13 Other half of Bit15
 	//                          //Bit12
 	U1STAbits.UTXBRK = 0;       //Bit11 Disabled
+#if (BOARD_TYPE == AUAV3_BOARD)
+	// Prevent issue 17 of processor silicon errata
+	// Ensure that other STA register bits are not changed until BRK hardware has cleared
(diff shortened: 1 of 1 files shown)


phollands:
Microchip are a little vague in their recommended solution for this issue. So I've done my own.  The dis-assembly of the code is shown below:-
```

67:                	//                          //Bit12
68:                	U1STAbits.UTXBRK = 0;       //Bit11 Disabled
007DA0  A96223     BCLR 0x223, #3
69:                #if (BOARD_TYPE == AUAV3_BOARD)
70:                	// Prevent issue 17 of processor silicon errata
71:                	// Ensure that other STA register bits are not changed until BRK hardware has cleared
72:                	while(U2STAbits.UTXBRK == 1);
007DA2  000000     NOP
007DA4  801195     MOV U2STA, W5
007DA6  208004     MOV #0x800, W4
007DA8  628204     AND W5, W4, W4
007DAA  520FE0     SUB W4, #0x0, [W15]
007DAC  3AFFFB     BRA NZ, 0x7DA4
73:                #endif

```

phollands:
On Re-reading issue 17 carefully, I can see that the bug only applies if the code is sending BRK characters down the serial stream. The silicon bug only occurs when clearing the BRK bit from the STA register (turning the serial break off). Since we never use or set the BRK, the bug will not be exposed. I also know through testing that this errata is not the cause of our other issue associated with intermittent non startup of the telemetry to the OpenLog. So I am closing this Pull Request.

PR #43, opened 2016-08-26
Prevent warning that ini_read() and ini_rename are "defined but not used"

phollands:
This change is being made to remove a compiler warning in the project that ini_read() and ini_rename are being defined but not used. 

Here is what was happening:-
minIni.c included miniIni.h  which, in turn,  included minGlue-mdd.h. 
minGlue_mdd.h has two statically defined functions called ini_read() and ini_rename.
This all works fine for minIni.c which uses ini_read() and ini_rename().

However, config.c, also include minIni.h, but will never directly use ini_read() and ini_rename(). Because these are are statically defined functions within the old include structure,  they are separately instantiated inside of config.c (in a separate name space), and so are created inside of config.c and never used.

This commit, ensures that the functions are not defined in config.c.  minGlue-mdd.h is no longer included  within miniIni.h.

Robert, if you are reading this, I would be grateful if you could test this fix for non microchip platforms (if you have been running this on them.). e.g. Unix and Windows.

Best wishes, Pete

### DIFF
--- a/MatrixPilot/MatrixPilot-auav3.X/nbproject/configurations.xml
+++ b/MatrixPilot/MatrixPilot-auav3.X/nbproject/configurations.xml
@@ -1,24 +1,25 @@
-<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <configurationDescriptor version="62">
-  <logicalFolder name="root" displayName="root"
-  projectFiles="true">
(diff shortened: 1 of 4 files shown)


phollands:
I'm going to withdraw this pull request. Unfortunately I've included a configurations.xml file by mistake in the commit, and I can see no way of "rebasing" the branch to remove it. So I'll re-do this on another Pull Request.

PR #51, opened 2016-09-13
Request git to ignore compiler generated binaries and transitory files

phollands:
I am adding a .gitignore text file which specifies which directories and file types that can be ignored for source code control purposes. This makes it much easier to use the git command line based tools, and  also makes the windows in sourcetree (a git GUI) cleaner.


wpremerlani:
Pete,
Good idea. Thanks.
Bill

PR #52, opened 2016-09-14
Fix linker/loader flag in X16 toolchain makefile

phollands:
This is a small but significant fix for the X16 compiler options when using the build-all.sh script to generate multiple  hex images for multiple target architectures (UDB4, UDB5, AUAV3 and SIL).

The script was failing to link correctly before this commit.


Inspirati:
Brilliant - nice find Peter. I believe this'll be the bug that prevented me
from using the latest compiler with the travis build system (-:

PR #53, opened 2016-09-17
Move common parameters for flight plans in  to options.h.

### LINKED ISSUE #33: WAYPOINT_RADIUS is redefined
I have this warning when I compile master brach

```
In file included from ../../MatrixPilot/flightplan-waypoints.c:44:0:
../../MatrixPilot/waypoints.h:26:0: warning: "WAYPOINT_RADIUS" redefined
../../MatrixPilot/flightplan.h:28:0: note: this is the location of the previous definition
```

phollands:
In current master, WAYPOINT_RADIUS, USE_FIXED_ORIGIN, and FIXED_ORIGIN_LOCATION are common to both the Logo and Waypoints.h flight plan types. The code compiles both flight plan types at all times nowadays. (A move made by Robert who I believe had a vision parameters to be dynamically configurable without a re-compile). Because all code is compiled, these parameters can no longer be held in both flightplans-logo.h and wayoints.h or they will cause a warning from the compiler about parameters being re-defined.

This commit simply moves the common parameters from flightplan.h to options.h to make them more visible.

I have changed WAYPOINT_RADIUS to WAYPOINT_PROXIMITY_RADIUS. Many people who read the phrase WAYPOINT_RADIUS wrongly believe that this will in some way be the radius which the plane will fly around a waypoint, when in fact, it is the distance from the waypoint finish line at which the plane will deem itself to have arrived. I think that WAYPOINT_PROXIMITY_RADIUS will make this clearer.

In a second later commit, I would recommend that we move the flightplan-logo.h flight plan, and the waypoints.h flight plans to the Config directory.

This commit  fixes #33 issue.

This preserves Robert's work, while dealing with one of my concerns which was that these common parameters were becoming rather hidden. Comments encouraged and welcomed on this request.


phollands:
Flight Tested OK. Works correctly in setting USE_FIXED_ORIGIN and ORIGIN_LOCATION along with the subsequent display of waypoints by flan.pyw in Google Earth.

So I am just awaiting any further comments before merging this into master.

PR #55, opened 2016-09-20
Fix for barometer code

KeesGuijt:
-add missing include estAltitude.h
-fix bug with 4Hz Gps modules
-allow for using USE_PRESSURE_ALT combined with USE_EXTENDED_NAV
-remove incomplete agl code
-repair commented out code in telemetry.c

### DIFF
--- a/libDCM/estLocation.c
+++ b/libDCM/estLocation.c
@@ -33,7 +33,6 @@ static void location_plane(int32_t* location)
 {
 	location[1] = ((lat_gps.WW - lat_origin.WW)/90); // in meters, range is about 20 miles
 	location[0] = long_scale((lon_gps.WW - lon_origin.WW)/90, cos_lat);
-//	location[2] = (alt_sl_gps.WW - alt_origin.WW)/100; // height in meters
 #ifdef USE_PRESSURE_ALT
 #warning "using pressure altitude instead of GPS altitude"
 	// division by 100 implies alt_origin is in centimeters; not documented elsewhere
--- a/libDCM/estLocation.c
+++ b/libDCM/estLocation.c
@@ -23,6 +23,7 @@
 #include "gpsData.h"
 #include "gpsParseCommon.h"
 #include "estLocation.h"
+#include "estAltitude.h"  //USE_PRESSURE_ALT
 #include "mathlibNAV.h"
 #include "estWind.h"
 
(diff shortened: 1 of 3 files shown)


libDCM/estLocation.c
```diff
@@ -32,7 +33,16 @@ static void location_plane(int32_t* location)
...
+#warning "using pressure altitude instead of GPS altitude"
+	// division by 100 implies alt_origin is in centimeters; not documented elsewhere
+	// longword result = (longword/10 - longword)/100 : range
+	location[2] = ((get_barometer_altitude()/10) - alt_origin.WW)/100; // height in meters
+#else
 	location[2] = (alt_sl_gps.WW - alt_origin.WW)/100; // height in meters
+#endif // USE_PRESSURE_ALT
+
 }
 #else // !USE_EXTENDED_NAV
```
phollands: What happens if the user is using the barometer without EXTENDED_NAVIGATION ? Is there any reason for keeping that option; if EXTENDED NAVIGATION works, then should we just move over to using it all the time ? (A question for Bill really).

KeesGuijt:
Hi, 
Yes, I'm aware of that. The comment lines where already there, i just added an indent. I'm ok with removing the commented lines. The comments in telemetry.c seem useful to me, i suggest we leave them there.

I've been using EXTENDED NAVIGATION for a few months now, without a problem. I read a comment by Bill some time ago, saying he believed EXTENDED NAVIGATION was already on by default.

Regards, Kees

phollands:
I just checked and USE_EXTENDED_NAV is [commented out in the options.h of Master](https://github.com/MatrixPilot/MatrixPilot/blob/master/Config/options.h#L778).  Perhaps, we should default it to on.

Bill, have you some thoughts on this ? Is it well tested, and should we default it to on ?

wpremerlani:
Peter,

Extended navigation has been exhaustively tested by Florin using HILSIM, he
has done many extremely long flights (36 hours in a Boeing 747).

Florin has also done extremely long actual flights when he drops his glider
at 150,000 feet altitude, and far down wind, from a balloon.

Extended navigation has nothing to do with the barometer, that is an
entirely separate thing. Extended navigation has to do only with doing 32
bit waypoint navigation arithmetic. I have not tested the barometer code,
and neither has Florin, with or without extended navigation.

I would suppose that you could use them together, but you could also very
well want to use extended navigation, but not the barometer.

Best regards,
Bill

KeesGuijt:
Bill, Pete,

Thanks for merging this fix.
Both combinations are suppported now, and i've tested both. So i think we should set Extended nav as default, regardless the barometer option.

Best wishes,
Kees

PR #58, opened 2016-09-22
fix negative headings in .csv file produced by flan

KeesGuijt:


phollands:
Kees,

entry.heading_degrees is a floating point number calculated within flan.pyw. So it is possible to have a heading of 359.99. For this reason, I think I would prefer a small change so that the code is:-
 ```
            if (entry.heading_degrees < 0 ):
                entry.heading_degrees +=360
            if (entry.heading_degrees >= 360 ):
                entry.heading_degrees -=360       
```        
This is what I've been using in my flan.pyw of late.  Do you agree with this refinement ?

Best wishes, Pete

KeesGuijt:
You are so right, nice!

Issue #59, opened 2016-09-22
type warning when selecting Hilsim + GPS_UBX_4HZ


KeesGuijt:
Hi,

I get this warning :
../../libDCM/gpsParseUBX.c: In function 'msg_PL1':
../../libDCM/gpsParseUBX.c:628:7: warning: format '%u' expects type 'unsigned int', but argument 3 has type 'unsigned int'

Kees

PR #60, opened 2016-09-22
Fix warning from compiler when using HILSIM from Ublox Parser.

phollands:
Kees,

This is a fix for #59 .

payloadlength is a signed integer in the[ ublox code file.](https://github.com/MatrixPilot/MatrixPilot/blob/master/libDCM/gpsParseUBX.c#L40)

I had to change [both variables in the printf](https://github.com/MatrixPilot/MatrixPilot/blob/master/libDCM/gpsParseUBX.c#L628) statement to remove the warning.

Best wishes, Pete

### DIFF
--- a/libDCM/gpsParseUBX.c
+++ b/libDCM/gpsParseUBX.c
@@ -625,7 +625,7 @@ static void msg_PL1(uint8_t gpschar)
 					}
 					else
 					{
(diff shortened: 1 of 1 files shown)


KeesGuijt:
Pete,

That was fast. The warning is gone after this change.

Thanks.
Kees

phollands:
The warning had been bugging me. I like a clean compile of Matrixpilot
these days.

PR #61, opened 2016-09-23
Simplify barometer configuration

KeesGuijt:
This aims to make configuring barometer easier. This means that enabling but not using barometric pressure is no longer possible.
rename  BAROMETER_ALTITUDE to USE_BAROMETER_ALTITUDE;
the macro BAROMETER_ALTITUDE might suggest an altitude value is configured, which is not the case
USE_BAROMETER_ALTITUDE makes clear this is a 0/1 setting.
remove #define USE_PRESSURE_ALT because it is not obvious that is should be commented out, not set 0 to inactivate.
also, a commented line would need to be added to indicate how to enable it, we don't like that.
LAUNCH_ALTITUDE is not needed anymore, because it is read from GPS automatically when the calibrations are performed (double waggle)

### DIFF
--- a/Config/options.h
+++ b/Config/options.h
@@ -148,15 +148,9 @@
 // the magnetometerOptions.h file, including declination and magnetometer type.
 #define MAG_YAW_DRIFT                       0
 
-// Define BAROMETER_ALTITUDE to be 1 to use barometer for altitude correction.
+// Define USE_BAROMETER_ALTITUDE to be 1 to use barometer for altitude correction.
 // Otherwise, if set to 0 only the GPS will be used.
-// If you select this option, you also need to correctly set the LAUNCH_ALTITUDE
-// to your takeoff location altitude at the time of initialisation.
-#define BAROMETER_ALTITUDE                  0
-
-// Set your takeoff/launch/initialisation altitude in meters.
-#define LAUNCH_ALTITUDE                     300
-
+#define USE_BAROMETER_ALTITUDE              0
 
 // Racing Mode
 // Setting RACING_MODE to 1 will keep the plane at a set throttle value while in waypoint mode.
--- a/MatrixPilot/console.c
+++ b/MatrixPilot/console.c
@@ -114,11 +114,10 @@ static void cmd_adc(char* arg)
 static void cmd_barom(char* arg)
 {
 #if (SILSIM != 1 && PX4 != 1)
-	printf("Barometer temp %i, pres %u, alt %u, agl %u\r\n",
+	printf("Barometer temp %i, pres %u, alt %u\r\n",
 	       get_barometer_temperature(),
 	       (uint16_t)get_barometer_pressure(),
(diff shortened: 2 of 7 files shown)


phollands:
I recently took off in fly by wire, but had trouble flying low in fly by wire. On analysis, the origin altitude had been set by the GPS at 40 meters above the actual height. I was at 77 meters ASL, and the origin was set at 117 meters.  The Satellites in View was only 4, which later moved up to 6 in flight. So on more recent flights I have reverted to setting the absolute origin (now moved to options.h where it should always have been).

I note that the [initialisation of the barometer will pickup the origin altitude automatically](https://github.com/phollands/MatrixPilot/blob/master/libDCM/estAltitude.c#L51), even if it is using an Absolute origin altitude and position. So that is good. That effectively replaces LAUNCH_ALTITUDE. So user's have a choice of whether to set the ASL using the Absolute Origin, or trust to a a good GPS reception at boot time, and accept the risk of a poor estimate.

wpremerlani:
Peter,
Which GPS were you using?
I found that the EM506 tends to declare a valid nav solution prematurely,
and that there is significant altitude estimation error for a while.
That is why shortly after the EM506 came out, I increased the intentional
delay before recording the origin. Based on tests that I did at the time, I
thought the delay was large enough, but apparently it is not enough in all
cases. So, it is a good idea to use a fixed origin.

Also, do you think we need to further increase the intentional delay? Or
perhaps also wait for a minimum number of satellites?

Best regards,
Bill

KeesGuijt:
gents, 

Comparing GPS with barometer, i often see GPS cycling 15m up, 0, 15m down in one to two minutes.

For me using barometer, altitude is allways 0, also after the flight. If the weather is not changing too much, barometer beats GPS altitude hands down.  

All the best,
Kees

chiloschista:
Hello ensemble,
please don't forget that GPS, geomagnetism and radio signals in general can be affected by sun activity. A good practise is to check sun activity and satellites position on the web, e.g. 
http://www.n3kl.org/sun/noaa.html
Here you can see sat prediction for today at your location

http://satpredictor2.deere.com/chartvisibility?addr=&lat=51.4756592&lon=-1.1047880999999506&timeZ=17&evMask=10&gps=YES&glonass=NO

http://satpredictor2.deere.com/chartdop?addr=&lat=51.4756592&lon=-1.1047880999999506&timeZ=17&evMask=10.0&gps=YES&glonass=NO

You can notice a couple of bad moments.
Best regards,
Ric

phollands:
Bill,

I took off around 1837GMT (1737BST) on 17th September (according to the
inflight video), so the initialisation was just before that; and that puts
me right in a Yellow spot of your graph with only 6 GPS satellites
available initially, before it then went up to 7, which is also the maximum
Satellites in view recorded in my telemetry. So a low number of sats a boot
up time,  was a large part of it. At times I have as many as 10 satellites
with an HDOP of 2.

Thanks for the link.

Pete

PR #64, opened 2016-09-24
Add desiredHeight to SERIAL_UDB_EXTRA telemetry,

phollands:
and add desired height to csv file produced by flight analyzer (flan.pyw).

This feature will allow Leo to display the setpoint of commanded altitude from the throttle channel, on the overlay of his video using Dashware.  I also enclose a sample graph from libreOffice showing the relationship between IMU Z altitude and desiredHeight from a flight today.

desired_height

I am wary of adding data to the SERIAL_UDB_EXTRA F2: format statement, as on older cards, the time to store data on the memory card can allow the OpenLog buffer to overrun (no flow control). So I'm open to suggestions whether we go ahead with this change. It works on my OpenLog at 57600 baud.

### DIFF
--- a/MatrixPilot/telemetry.c
+++ b/MatrixPilot/telemetry.c
@@ -29,6 +29,7 @@
 #include "cameraCntrl.h"
 #include "flightplan.h"
 #include "flightplan_waypoints.h"
+#include "altitudeCntrl.h"
 #if (USE_TELELOG == 1)
 #include "telemetry_log.h"
 #endif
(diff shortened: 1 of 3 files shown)


phollands:
Bill,

I went and dug out the log file for flight 00014, and the GPS fluctuated  a huge amount, which I've never seen before, while the plane was on the ground.  My ASL was 77 meters, the GPS went from over 200 meters to -100 meters before the flight. I enclose a graph. I have never seen that happen before to that extent.

fluctuating_gps

phollands:
Apologies: Ive just written to Bill on the wrong Pull Request. This note should have been on #61

PR #65, opened 2016-09-26
Update MAVLink generator, MAVProxy from upstream repositories and generate C include files

phollands:
It's been well over a year since we have updated the core code for MAVLink from the main github.com mavlink repositories. The same is true for MAVProxy.  

An overview of the process used in this Pull Request to update and re-generate mavlink is as follows:_

1) I forked the main mavlink and MAVProxy code from their respective official repositories into the [MatrixPilot account on github.](https://github.com/MatrixPilot?tab=repositories)

2) I replaced the existing directoriess in Tools/MAVLink/pymavlink and Tools/MAVLink/MAVProxy with git submodules that point back to the MarixPilot account on github. This will make it easier to pull updates from upstream, and to provide commits back upstream as well. Their projects are embedded as additional git repositories within our repositories.

This Pull Request is a pre-cursor to another yet to be created PR to  update the SERIAL_UDB_EXTRA (SUE) over mavlink messages, and update  flan.pyw, to provide full legacy support to those people who are ready to move over to mavlink, but would like to ensure their telemetry still has all the specific MatrixPilot telemetry messages that they have had with SUE.

I have some in depth testing to do on this PR before it will be merged, but would like people to see what I'm doing, before I am too far into  flight testing.

Best wishes, Pete


phollands:
I am minded to close this Pull Request.

Although the updated mavlink code is working, the use of git submodules is causing me grief and delays. So I am considering going back to just including all these remote projects in our distribution of matrixpilot in the way that we have in the past.

My issues with the git submodules for MatrixPilot at the moment are:
*  developers and users have to know to use  "git clone --recursive" or other special commands to ensure they include the submodules when they download MatrixPilot. 
* Switching branches in the MatrixPilot repository is broken after using git submodules. Something is getting broken in the git database. I am using sourctree and pure git commands. That looks bad. 
* flan.pyw is going to need the MatrixPIlot specific python dialect, which currently lives in a location that is in a submodule. So everyone would need submodules to work correctly, and to be able download them.

So, all in all, given that we only do this update about once a year, I think I'm going to remove the idea of submodules, and just include all the necessary code in our own MatrixPilot git repository. This is how I updated mavlink last time.

phollands:
After reading [this blog entry on git submodules,](https://medium.com/@porteneuve/mastering-git-submodules-34c65e940407#.l4t2lfqja) I have decided to drop them, and start again.

Issue #75, opened 2016-10-10
Activating OSD_NATIVE on UDB5 board broke GPS Lock


elgarbe:
When I activate OSD_NATIVE on my UDB5 board, broke GPS signal. It is tested, and don't work, in bit bashing mode. 
OSD seems to work ok becouse I can see artificial horizon update when I move my plane.

phollands:
I just want to note here in this issue log, that there was some [prior discussion of the issue on uavdevboard at this link.](https://groups.google.com/forum/#!topic/uavdevboard/iOtT4jfuu-8).

I did some tests this morning where I enabled the OSD with legacy pinout, and then commented out the main routine called [mp_osd_run_step() ](https://github.com/phollands/MatrixPilot/blob/inverted_test/MatrixPilot/mp_osd.c#L554)which updates the OSD with the latest telemetry. GPS Lock was obtained in about a minute. I then took out the commenting so that mp_osd_run_step() was live and working. I rebooted. The GPS lock was **not** obtained after 5 minutes of the plane running MatrixPilot. I rebooted again, and GPS lock again was not obtained after 5 minutes of running the code.

You can see my [test branch at this location](https://github.com/phollands/MatrixPilot/commits/inverted_test), if you want to understand what I've been doing in more detail. (That link may go stale when I next rebase the branch onto any new changes in master).

phollands:
If I comment out 1/4 of the main meat of the mp_osd_run_step() code, then my orange led flashes (rx interrupt on gps toggled on each interrupt) and my blue led flashes (full completion of gps parsing of messages, ready to commit lat / long etc). 

If  instead I keep all the mp_osd_run_step() code in place, but reduce the frequency of calling it from 8hz to 2HZ, again my blue and orange test LEDs start flashing correctly.

So I am starting to think that somehow, the priorities of the interrupts in the code are potentially not correct (or not what we think they are), and that the heartbeat_pulse() code which should be at priority 3, is actually at the same or a higher interrupt priority than the serial IO rx for the gps, which should be higher and be set at priority 4. In theory, there is no way that the mp_osd_run_step() should be able to stop the serial IO RX interrupt from being fired but that is what the tests say is happening. 

Full details of each commit, and the results, for my testing is at [my inverted_test branch.](https://github.com/phollands/MatrixPilot/commits/inverted_test)

wpremerlani:
Peter,

It sounds like you are closing in on the issue. If the OSD is using "bit
[...]

PR #77, opened 2016-10-11
Fix GPS initialisation when OSD SPI bitbash used with AUAV3 or UDB5,

phollands:
Split udb_heartbeat_counter and create udb_pulse_counter for low priority heartbeat.

This is a fix for issue #75 in which activating the OSD with the UDB5 using SPI bit bashing, when using the UDB5, would result in the GPS (EM506 or Ublox) not getting initialised correctly. It is expected that the issue also effected the AUAV3 because they both use a 200HZ Heartbeat, unlike the UDB4 which uses a 40HZ heartbeat.

The root cause was that udb_heartbeat_counter was incremented in a high priority interrupt (heartbeat()), but was being used in a low priority interrupt (heartbeat_pulse). If heartbeat_pulse took more than 1/200th of a second to complete, then udb_heartbeat_counter could get incremented while heartbeat_pulse() was part of the way through processing it's routines. See issue #75 for further details.

I will be flight testing it on the UDB5, with OSI bit bashing enabled, although I won't have the OSD hardware attached. Leo, it would be good if you could do a ground test of the OSD with the OSD hardware attached, and let me know the result. I think we need a full hardware test before approving this commit.

Best wishes, Pete

### DIFF
--- a/MatrixPilot/servoPrepare.c
+++ b/MatrixPilot/servoPrepare.c
@@ -83,7 +83,7 @@ void servoPrepare_init(void) // initialize the PWM
 
 static void flight_controller(void)
 {
-	if (udb_heartbeat_counter % (HEARTBEAT_HZ/40) == 0)
+	if (udb_pulse_counter % (HEARTBEAT_HZ/40) == 0)
 	{
 		flight_mode_switch_2pos_poll(); // we always want this called at 40Hz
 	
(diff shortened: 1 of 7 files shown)


phollands:
Well Done Travis. Thank you Robert for getting Travis working. [Travis tests failed](https://travis-ci.org/MatrixPilot/MatrixPilot/builds/166789038) for the software in the loop simulation, because my re-factoring was done with MPLAB-X which does not cover off code that is for non-microchip platforms like Software in the Loop.  I will go look into what needs to be done for SIL.

phollands:
Flight Test passed OK with OSD enabled and bit bashing enabled. Perfect flight. CPU usage averaged at 31% on the UDB5.

elgarbe:
I hope tomorrow morning could made a test flight with the OSD. If so, then I'll try to make a video of the OSD working on my ground PC.
But it seems that I will have bad wether for the rest of the week :-(

PR #78, opened 2016-10-14
change estimatedWind variable from array to vector type - no binary c…

Inspirati:
…hange
- working towards reducing the arbitrary differences between master and development branches.

### DIFF
--- a/MatrixPilot/MAVUDBExtra.c
+++ b/MatrixPilot/MAVUDBExtra.c
@@ -127,7 +127,7 @@ void MAVUDBExtraOutput_40hz(void)
 					    lat_gps.WW, lon_gps.WW, alt_sl_gps.WW, waypointIndex,
 					    rmat[0], rmat[1], rmat[2], rmat[3], rmat[4], rmat[5], rmat[6], rmat[7], rmat[8],
 					    (uint16_t) cog_gps.BB, sog_gps.BB, (uint16_t) udb_cpu_load(), voltage_milis.BB,
-					    air_speed_3DIMU, estimatedWind[0], estimatedWind[1], estimatedWind[2],
+					    air_speed_3DIMU, estimatedWind.x, estimatedWind.y, estimatedWind.z,
 					    magFieldEarth[0], magFieldEarth[1], magFieldEarth[2],
 					    svs, hdop);
 #else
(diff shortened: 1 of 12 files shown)


wpremerlani:
I have no strong objections to the change, but I was not aware there was a difference.
In any case, I prefer array over vector, because vector precludes the use of the matrxi/vector library.
If I had to do it over again, I would make all 2 and 3 dimension vectors arrays.

phollands:
Robert, 

Sounds like we should keep vectors as arrays so that we can matrix operate ( e.g. ["dot" product](https://en.wikipedia.org/wiki/Dot_product) and "cross" product them ) whenever that becomes useful.

One of the reasons Bill chose the dspic was it's strong low level hardware support for matrix operations, which is particularly important for code in libDCM which these days is running at 200Hz for the UDB5 and AUAV3. 

For code readability we could  replace  numeric indices with named indices, as in 
`estimatedWind[0]` with `estimatedWind[_X_] ` or something along those lines. I used _X_ to ensure the macro pre-processor does not replace some other X that happens to be in the code.

As this is an open source project which wants to help new developers,  I am all for making the code as readable as possible, especially when it has no impact on performance.

Best wishes, Pete

Inspirati:
I guess I am of a fundamental difference of opinion with BillP on this issue. I'd prefer to see the arrays replaced with truly functional types where ever possible, and where ever not, look forward to adjusting the supporting libraries to suit, rather than the other way around. If we could go with me on the flow on this one for the moment, at least with regards to this small change, it could serve well towards building the road to illustrating my point. I guess what is missing is my largely unpublished work with a regression testing suite..

Whilst I'm not proposing that we try and make this a C++ codebase, I am an advocate for adopting the some of the more significant principles that the OO methodology provides, and this is just a small example of how I'm trying to gently achieve such.

phollands:
Robert, if we are going to have a vector type, then we need to ensure that our low level routines can conduct linear algebra on the new vector type (as opposed to the arrays that they have been using previously).   So if we want to move in this direction, I think we do need to create a PR, which changes all relevant arrays into vector types, and **also** ensures that the supporting fast code for linear algebra supports the new format of vector types.  I have not looked yet at how the new vector type is represented in memory, and it may be that it is already supported. But the key point is that the vector type needs to be supported by fast linear algebra routines in MatrixPilot. (e.g. dot products and the like). Best wishes, Pete

PR #79, opened 2016-10-14
expand the travis automated build system to include more MatrixPilot …

Inspirati:
…support targets

### DIFF
--- a/Tools/FlashOSD/Config/options_file_system.h
+++ b/Tools/FlashOSD/Config/options_file_system.h
@@ -1,0 +1,238 @@
+// This file is part of MatrixPilot.
+//
+//    http://code.google.com/p/gentlenav/
+//
+// Copyright 2009-2011 MatrixPilot Team
+// See the AUTHORS.TXT file for a list of authors of MatrixPilot.
+//
+// MatrixPilot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// MatrixPilot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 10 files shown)


phollands:
Robert,

Thanks for your travis update, which looks fine to me.

I've pulled your branch, and get a couple of warning on compiles, which, when I reviewed them carefully have highlighted a couple of issues in the FlashOSD project which are not associated with your changes.

screen shot 2016-10-14 at 11 24 51

The  options_osd.h in the Tools/FlashOSD/Config directory has the wrong number of OSD_NONE as 1, when it should be 0, and instead OSD_NATIVE should be enabled instead of OSD_NONE. 

```
// define the OSD types
#define OSD_NONE            1   // OSD disabled
#define OSD_NATIVE          1   // native OSD
#define OSD_REMZIBI         2   // Output data formatted to use as input to a Remzibi OSD
#define OSD_MINIM           3   // Output data formatted for minim OSD
```
and also further down the options_osd.h file it is missing a line....
```
// OSD_NATIVE development options:
#define USE_OSD_SPI     0   // set this to 1 to use the SPI peripheral, 0 to bit-bash
#define OSD_SF          5   // scale factor for SPI delays - TODO: get rid of
```
now has and additional line as in the options_osd.h for the MatrixPilot project:-
```
// OSD_NATIVE development options:
// Set OSD_LEGACY_SPI_PINOUT to 1 if you wired up your UDB4 or UDB5 for the Native OSD
// before October 2016; In this case, The wiring does not follow the labels on the UDB boards.
// Otherwise, leave OSD_LEGACY_SPI_PINOUT_UDB as 0, for the pin out labels on the UDB boards
// to agree with the SPI signals of the OSD.
#define OSD_LEGACY_SPI_PINOUT   0   // Default setting is 0
#define USE_OSD_SPI             0   // set this to 1 to use the SPI peripheral, 0 to bit-bash
#define OSD_SF                  5   // scale factor for SPI delays - TODO: get rid of
```

I will review other projects like RollPitchYaw and LedTest in the light of recent changes. It's good to know that Travis will be compiling those in the future.

Best wishes, Pete

phollands:
I've updated FlashOSD options_osd.h in MatrixPilot/MatrixPIlot directly[ with this commit](https://github.com/MatrixPilot/MatrixPilot/commit/f69927cc0d3cc8b4f756842af7f9c62092d954f2).

PR #80, opened 2016-10-14
Telemetry.c tidy up

phollands:
Before starting on SUE over MAVLink, I would like to tidy up telemetry.c, mainly because it is hard to read with so many #if statement to include or exclude code.

I have taken out SERIAL_UDB as I am only aware of people using SERIAL_UDB_EXTRA. 
I have taken out SERIAL_UDB_MAG, as debugging of the magnetometer is usually done with SERIAL_MAGNETOMETER. 

SERIAL_MAGNETOMTER was printing out a variable called I2ERROR. If anyone can point me to where I2ERROR is defined, I would be grateful, as I cannot currently declare it correctly, and so have temporarily commented it out. 

I have some testing to do on this, but thought I would put these ideas past you all, before I go too far, if I need to back off.

Best wishes, Pete

### DIFF
--- a/Config/options.h
+++ b/Config/options.h
@@ -340,8 +340,8 @@
 
 
 ////////////////////////////////////////////////////////////////////////////////
-// Serial Output Format (Can be SERIAL_NONE, SERIAL_DEBUG, SERIAL_ARDUSTATION, SERIAL_UDB,
-// SERIAL_UDB_EXTRA,SERIAL_MAVLINK, SERIAL_CAM_TRACK, SERIAL_OSD_REMZIBI, or SERIAL_UDB_MAG)
+// Serial Output Format (Can be SERIAL_NONE, SERIAL_DEBUG, SERIAL_ARDUSTATION,
+// SERIAL_UDB_EXTRA,SERIAL_MAVLINK, SERIAL_CAM_TRACK, SERIAL_OSD_REMZIBI, SERIAL_MAGNETOMETER)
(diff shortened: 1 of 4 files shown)


wpremerlani:
Peter,
Regarding I2ERROR in the magnetometer debugging code, I do not recall what that was for, I think it was the status register of the communications to the magnetometer. In any case, go ahead and comment it out for now. Most likely it will never be needed again, but if it does, I can track it down then.
Best regards,
Bill

phollands:
This PR test compiled with all SERIAL_XXX options with UDB5. Telemetry output tested for SERIAL_NONE, SERIAL_DEBUG, SERIAL_ARDUSTATION, SERIAL_UDB_EXTRA. Flight tested with SERIAL_UDB_EXTRA.

Robert, if its of interest to you, then it would be good to have some additional travis tests for SERIAL_UDB_EXTRA and SERIAL_MAVLINK.  I'm thinking that the Config directory should have 3 subdirectories with different Configs: 

* "examples" (real world RC plane conigs from our core flyers that have flown with master),
*  "tests"  ( used for travis ),
*  simulation (I would personally like to see the Easy Star simulation all setup and easy to run; this is where and updated Cessna config would go.). 

My main point is  it would be nice, some time, to get the options in telemetry test compiled by travis.

If there are no further comments, I intend to merge this update to telemetry tomorrow.

Best wishes, Pete

PR #82, opened 2016-10-15
demonstrate using some build script options

Inspirati:


phollands:
Rob, Thanks very much, Pete

PR #85, opened 2016-10-15
Fix filesys opts

Inspirati:
revert rename of FSconfg.h to option_file_system.h since it does not belong as a user configurable header file. FSconfig.h needs to be removed anyhow in favour of using the FatFs library for filesys support.


phollands:
Hi Robert,

Please can you  change the project configurations MPLAB-X as well.

Currently, with this PR, the project gives errors on loading up for example MatrixPilot-UDB5. I find that if there are errors on loading, then the dynamic parser in the IDE can get itself in a muddle, so I have purposely removed such errors, and would like to keep it that way. You will need to check each of the relevant projects tha thad the options_file_system.h .

screen shot 2016-10-15 at 20 16 27

screen shot 2016-10-15 at 20 17 03

And yes, the sooner we expel the non-gpl v3 code from MatrixPilot the better.

I took a look at the vector library, which has a non GPL v3 license,  which is using dspic specific instructions sets. In other words, it can only ever work on a dspic anyway ... that is the point of the fast vector library. So I was going to look at whether it really needs re-coding independently of Microchip. It is nice that Bill has offered to recode it.

Thanks, Pete

Inspirati:
do you foresee any issues with the projects being updated with output from
the project generator script?

phollands:
You can always run it, and make a separate commit herein your branch fix-filesys-opts. 

 I'll test it. If it looks no good, then you can interactively rebase the branch for this PR, and take the commit out again.  I think there are 12 main compiles to test. (3 boards, 4 projects).

So please, go ahead, and add a commit using the generation script, and I'll do some test with it tomorrow. 

And yes, I think I did try it out, and that there were issues for me, but then I'm not so familiar with  how that all works.

As you can imagine, it has taken a lot of work, and manual testing the MPLAX-IDE to review can correct all the projects by hand. As I am now close to MatrixPilot 5.0 release, I do want to mostly keep the projects working well without any errors.  I note that FlashOSD is fast to build, and does not include to many files.  It seems to me that projects like LedTest  or RollPitchYaw seem to compile many many files that are not actually needed. For me that is not too much of a problem as a typical compile takes 25 seconds. But for other people, a compile takes 5 minutes, and reducing compile time is probably beneficial for them.

I think you may find that some of the XML indents are different in the later versions of MPLAB. That then makes it hard to track real changes in the git repository, as so much of the file changes just with indents.

Anyway, I would like to see your project generation script work. It would be good not to have to go through all the manual work of setting things up. 

Best wishes, Pete

MatrixPilot/MatrixPilot-auav3.X/nbproject/configurations.xml
```diff
@@ -500,7 +499,6 @@
         <itemPath>../../MatrixPilot/servoMix.c</itemPath>
         <itemPath>../../MatrixPilot/servoPrepare.c</itemPath>
         <itemPath>../../MatrixPilot/sonarCntrl.c</itemPath>
-        <itemPath>../../MatrixPilot/stack.S</itemPath>
```
phollands: Look like stack.S is no longer included. Is that meant to be ?

phollands:
Rob,

Thanks very much for this.  I am looking forward to testing the projects loading into MPLAB-X today, and reviewing the changes made. It's much better to have a script putting this all together.

Best wishes, Pete

phollands:
Rob,

I notice that when one loads a project in MPLAB-X, and then closes the project (with no other modifications), that git then shows a series of changes to the associated configurations.xml file.
I think it is because some of the template files that you are using with your python script have got dos line endings, while the main body of lines produced by python do not. So the configurations.xml file is partly having dos based line endings and partly unix.  The unix line endings seem to be what MPLAB-X generates, so can change take the CR character out of the template files and re-generate ?

Here is more detail on the above starting with a screenshot of the changes git sees after a project load and save:-
screen shot 2016-10-17 at 09 07 04

And here is configurations.xml view in the vi editor straight from the repository for your branch fix-filesys-opts, without loading a project with MPLAB-X. You can see the dos line endings (^M) at the top of the file, and then they stop.

screen shot 2016-10-17 at 09 04 47

Best wishes, Pete

phollands:
Rob, I've finished my testing and checks. It was nice to see your script fix up mag_drift.c for some of the projects. And also I appreciated the speed up of compilers for some of the smaller projects.

So I think that once the dos endings are sorted, we are good to merge.

Best wishes, Pete

phollands:
That all works well now. Nice Tool for project generation. Merging. Thanks very much.  Pete

PR #87, opened 2016-10-15
suble move towards runtime only flightplan build configurations

Inspirati:
potential subtle implications may need further consideration, however i believe they'll be no practical problems - and it moves towards cleaning up the otherwise extraneous conditional compilation paths.

### DIFF
--- a/MatrixPilot/MAVMission.c
+++ b/MatrixPilot/MAVMission.c
@@ -391,7 +391,7 @@ boolean MAVMissionHandleMessage(mavlink_message_t* handle_msg)
 {
 	switch (handle_msg->msgid)
 	{
-#if (FLIGHT_PLAN_TYPE == FP_WAYPOINTS)
+//#if (FLIGHT_PLAN_TYPE == FP_WAYPOINTS)
 		case MAVLINK_MSG_ID_MISSION_REQUEST_LIST:
 			MissionRequestList(handle_msg);
 			break;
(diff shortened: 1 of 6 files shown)


phollands:
Rob,

I not sure about PRs that are "subtle" . :-).  
I think we need the bigger plan here, and then ideally we need a more feature complete PR.

The meta issue here is whether we are moving towards one compilation for all features (removing the extensive use of #if includes), so that we can  produce one binary, which can be configured at run time, rather than at compile time. 

My own view is that the commercial market has matured way to a point of complexity way beyond the resources of this project. When the AUAV3 was released, there was an attempt to focus on a market window for a small device which could be sold to the masses. That window of opportunity, to this project for a commercial board has passed. The PIXracer using pixhawk and ardupilot is the natural successor. This project is a DIY project primarily for people with developer experience. To some extent our goals should be KISS (Keep It Simple Stupid). And keeping the project fully open source is key to its survival. The simplicity allows developers to experiment. Conversely, this project does not have the resources to engineer a lot of "me too" features of a vastly more complex total system with a hugely well integrated ground control system, and ground configuration tool to create a product for the masses.  When it comes to flight configuration, for the most part, I really appreciate having all my config within the git repository. That has become hugely useful. But as I say, may be this needs more discussion to arrive at more consensus. 

If this PR is not urgent, I would rather let is stay for now, until we have sorted out the bigger picture of where we are heading. If you feel strongly about this PR going into the code, then I will look at it in more detail.

In the mean time, I am focussing on  getting SUE over MAVLink working; I really would like to get to the point of releasing MatrixPIlot 5.0 .

Best wishes, Pete

**sub·tle**
ˈsədl/
adjective
   * (especially of a change or distinction) so delicate or precise as to be difficult to analyze or describe.
   * his language expresses rich and subtle meanings"
synonyms:	fine, fine-drawn, nice, hair-splitting
   * "subtle distinctions"
   * (of a mixture or effect) delicately complex and understated.
   * "subtle lighting"
synonyms:	understated, muted, subdued; More
   * making use of clever and indirect methods to achieve something.
  * "he tried a more subtle approach"

phollands:
HI Robert, the last comment on this PR was by myself October 17th, more than a month ago. Should we continue with this PR, or do you think we should close it ? Best wishes, Pete

PR #88, opened 2016-10-16
Fix reference to flightplan-logo.h in UDB LOGO editor

elgarbe:
Fix reference to flightplan-logo.h in UDB LOGO editor to the new config directory


phollands:
Leo, Thanks very much for that fix. Best wishes, Pete

PR #90, opened 2016-10-19
Update  SERIAL UDB EXTRA (SUE) over MAVLINK

phollands:
The goal of SUE_over_mavlink is to allow pilots to use mavlink without losing any of the historical benefits of using SUE. It usually runs alongside the main common mavlink messages used all autopilots that support mavlink.

In general it is expected that the Mavlink packets are captured on the OpenLog in the same way as SUE ascii text. Both Mavlink and SUE will generate a LOG.TXT file. But the MAVLink LOG.TXT file is a binary file. When the LOG.TXT binary text file is selected using flan.pyw, it will offer to rename the file to LOG.RAW (raw mavlink file without GCS time stamps).

From  a flan.pyw point of view, once the mavlink messages are parsed, the normal python code is used to generate Google Earth KML files, csv files etc.  There is not difference.  But, for mavlink,  flan.pyw goes one stage, it also re-generates the LOG.TXT ascii file that would have originally been seen on the OpenLog.  So now the user has a binary LOG.RAW  file, and a legacy LOG.txt file.  The best of all worlds. 

The MAVLINk user, can connect his telemetry stream both to an OpenLog and an Xbee wireless transmitter at the same time. This allows PID's to be tuned in flight, when the aircraft is near the pilot, greatly speeding up the tuning process. But at long range, the OpenLog can be relied upon, to faithfully capture all the plane's transmitted telemetry, for further analysis, without any drop of packets.

The commits below should be fairly self explanatory. I have done an initial test flight of the mavlink code, and parsed the resulting OpenLog file. 

Some more testing will be required before I merge this, but would like to publish the work so far. 

As always, I would be delighted to receive any comments.

Best wishes, Pete


phollands:
Flight Tested OK.  
Google Earth KMZ OK.
CSV file in spreadsheet OK.
Ascii LOG.TXT looks OK on inspection.

phollands:
I have put in a [pull request](https://github.com/mavlink/mavlink/pull/640) to the upstream mavlink repository to accept our updated matrixpilot.xml specification of our messages.

This should mean that our SUE over Mavlink mesages will be, to some extent, understandable by third party mavlink based GCS system in the future. (As at least being able to inspect the messages).

magicrub:
Has the @MatrixPilot project considered not storing+committing the mavlink generated files? The ArduPilot project (and maybe PX4 too?) have moved to committing only the xml and generate on every compile. When there's no change in xml it skips through the already-generated files very fast.

PR #96, opened 2016-11-14
Added code to make OSD_NUM_SATS to flash when SVS is less than a configurable value

elgarbe:
unfortunately I mess up previouse PR trying to squash all the commits into one.
I put this new one after squashing the commits in my working copy


phollands:
Thanks Leo for this new feature. Merging.

PR #97, opened 2016-11-16
RSSI signal, in OSD, will blink when the signal strenght drops below a configurable value

elgarbe:
Just a blinking alarm on low RSSI in the native OSD
[Video for the new feature](https://youtu.be/rVJ379R7bZ0)

### DIFF
--- a/MatrixPilot/mp_osd.c
+++ b/MatrixPilot/mp_osd.c
@@ -337,6 +337,7 @@ static void osd_update_values_phase_1(void)
 	int8_t dir_to_goal;
 	int16_t dist_to_goal;
 	struct relative2D curHeading;
+    static char f_OSD_RSSI_flash = 0;        // To keep track on show/hide OSD_RSSI
 
 	curHeading.x = -rmat[1];
 	curHeading.y = rmat[4];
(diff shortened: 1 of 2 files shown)


chiloschista:
Hello,
I would like to have those blinking warnings selectable.
I'm not a fan of blinkers, they are distracting, I prefer keep them 
under control myself.
RSSI can jump a lot, especially with certain systems. It would then 
start and stop blinking all the time.
Not sure that 10% is a good value also, it depends where a user set his 
usable range. On several planes I did set the no trepassing level at 0%, 
knowing at 0% I still had control. It is not an absolute value.
Same could be sayd about satellites number.
Ric

Il 16.11.2016 13:55, elgarbe ha scritto:

elgarbe:
Ok, no problem. I will add conditional compilation and treshold configuration.
Is the word FLASH well used here? do I have to use blink instead?

Leo

phollands:
Flash or blink is fine for me.

elgarbe:
I rebase this branch onto new master and solve a conflict.
Now it is ok to merge.
Thank

phollands:
Ric, Thanks for your comments. Leo has made GPS snd RSSI "low value flashing" an optional feature. Merging. Best wishes, Pete

PR #98, opened 2016-11-21
Test of the divider sample_count before calling the __buitin_divsd

Jeffem:
To avoid an eventual divide by zero exception already encountered on
UDB5

### DIFF
--- a/libUDB/analog2digital_udb5.c
+++ b/libUDB/analog2digital_udb5.c
@@ -246,7 +246,8 @@ void __attribute__((__interrupt__,__no_auto_psv__)) _DMA0Interrupt(void)
 
 	// When there is a chance that data will be read soon,
 	// have the new average values ready.
-	if (sample_count > ALMOST_ENOUGH_SAMPLES)
+    //Modif gfm#1 to avoid an eventual divide by zero exception : add test sample_count>0
+	if ((sample_count > ALMOST_ENOUGH_SAMPLES) && (sample_count>0))
 	{
 		udb_vcc.value = __builtin_divsd(udb_vcc.sum, sample_count);
 		udb_5v.value  = __builtin_divsd(udb_5v.sum,  sample_count);


phollands:
Guy-Francois, Thanks for this. 

While reviewing this pull request, and checking how a divide by zero might happen,  I noticed that the __builtin_divsd is now a macro ! It has been changed from using the builtin hardware for the dspic to using a standard divide as in:-
`#define __builtin_divsd(x,y) (( int16_t)((( int32_t)(x))/(int16_t)(y)))`

This is now true for[ all __builtin functions, of any multiply or divide. ] (https://github.com/MatrixPilot/MatrixPilot/blob/f021fcd6d7f01d540dd0df2212f6a2a6d6007c61/libUDB/builtins.h) . This change was made in master in Jan 2015. 

Bill, did you know about that ? And Robert, do you want to explain the reason for this ? Was it for the PX4 work that you have been doing ? Can we not differentiate between a PX4 and a UDB4/5 and AUAV3 for these purposes ?

Best wishes Pete

phollands:
Guy-Francois,

As ALMOST_ENOUGH_SAMPLES is set to 216 for the UDB, the new additional statement that you have added ( && sample_count > 0 ) does no more than the original statement (if sample_count > ALMOST_ENOUGH_SAMPLES).  So I don't see that the additional statement can help remove the potential problem if it is still around.

In the mean time, I am also looking very carefully at how a divide_by_zero might have happened with the code as it stands. 

But it would be nice if we could make the problem happen again. It's always hard to fix a problem that does not occur.

Best wishes, Pete

wpremerlani:
Peter,
Regarding the macro for divsd, I thought it was used only for SILSIM. What
a shame if it is being used with the UDB5, we should fix that if that is
the case.
Most, if not all of the divides by integers were put there by me, and I
have gone to great pains to assure that divide by zero cannot happen. If
there are few still there, we should fix them. You have to special case the
situation in which the denominator is zero.
And I agree with your comments Peter:
"(if sample_count > ALMOST_ENOUGH_SAMPLES). So I don't see that the
additional statement can help remove the potential problem if it is still
around."
Best regards,
Bill

Jeffem:
Hi Pete,

I got the problem at the first use of my UDB5 card, when I try to run RollPitcYaw.

It was at the beginning of 2015.

I will try to re-open my archive to put the UDB in the same environment.

I am not sure to reproduce the pb, but I understand very well that if I can that will give you oportunity to go deeply in.

I will come back on that subject next week.

Best wishes, gfm

 

De : Pete Hollands [mailto:notifications@github.com] 
Envoyé : lundi 21 novembre 2016 18:14
À : MatrixPilot/MatrixPilot
Cc : Jeffem; Author
Objet : Re: [MatrixPilot/MatrixPilot] Test of the divider sample_count before calling the __buitin_divsd (#98)

 

Guy-Francois,

As ALMOST_ENOUGH_SAMPLES is set to 216 for the UDB, the new additional statement that you have added ( && sample_count > 0 ) does no more than the original statement (if sample_count > ALMOST_ENOUGH_SAMPLES). So I don't see that the additional statement can help remove the potential problem if it is still around.

In the mean time, I am also looking very carefully at how a divide_by_zero might have happened with the code as it stands.

But it would be nice if we could make the problem happen again. It's always hard to fix a problem that does not occur.

Best wishes, Pete

Inspirati:
For sure, this is just another example of some of my porting work being left in a state of ongoing development, and at some point then becoming lost in the fog. A dead give-away of this is the amount of commented out code options in the file you reference. However in order to make things right again would require a better test setup than is currently at my disposal, so I myself wouldn't dare to tackle changing it on my own right now. Please guys, feel free to fix this up as best you see a way how. Hopefully the automated build system will catch if something goes wrong with the SILSIM builds.

BTW; given the time frame of my commit, it was possibly due to my work at the time towards supporting a potential port to a PIC32 based board which Nikolay and I were once working on! Sorry to all for leaving it only half baked )-:

MatrixPilot:
Robert,  Well, it is very interesting that I unwittingly did all my flight tests over the summer with the __builtin_ macros and   all 3 boards .... and everything appeared to work just fine.  :-) . 

When I get a moment, I'll fix this back up, the way that Bill expected this to be done and make a PR for you  to review (specifically to check the other architectures still compile correctly). Bill originally chose the dspic specifically because of the hardware support for these fast divide and multiplies.

I expect the CPU usage will not show much change, but the speed of executing this high priority interrupt routine should go up considerably.

Best wishes, Pete

Inspirati:
Those of you who are interested in looking at the exception handling may be interested in my half-baked setjmp/longjmp implementation which is already in master.

However with very little detailed investigation of its effectiveness. IIRC, what i've implemented is hopefully somewhat preferable to what we had, which was exception handlers which simply went into a permanent noop loop..

At the time i figured trying to start again from main was at least a better option than definitely doing nothing more than waiting for debugger intervention!

Whilst the code currently does nothing in an effort to correct the cause, I did imagine the potential for a plane to keep flying, despite the likelihood of repeatedly pulling the same exception over and over again.

So this would be an example of choosing wishful thinking over certain death /-:

Inspirati:
I've got a very vague memory of looking at the performance difference, and
noting that it was not notable. However I'm all for using the hardware
features where-ever possible, so it'll be great to have this finally fixed
up.


Once again I suspect the commented out code shall be helpful as
documentation of the possible path to perfection:

#ifdef (UDB4) || (UDB5) || (AUAV3)
int32_t __builtin_mulss(int16_t p0, int16_t p1);
int32_t __builtin_mulus(uint16_t p0, int16_t p1);
int32_t __builtin_mulsu(int16_t p0, uint16_t p1);
uint32_t __builtin_muluu(uint16_t p0, uint16_t p1);
uint16_t __builtin_divud(uint32_t num, uint16_t den);
int16_t __builtin_divsd(int32_t num, int16_t den);
#else
#define __builtin_mulss(x,y) ((( int32_t)(x))*( int32_t)(y))
#define __builtin_mulus(x,y) (((uint32_t)(x))*( int32_t)(y))
#define __builtin_mulsu(x,y) ((( int32_t)(x))*(uint32_t)(y))
#define __builtin_muluu(x,y) (((uint32_t)(x))*(uint32_t)(y))
#define __builtin_divud(x,y) ((uint16_t)(((uint32_t)(x))/(uint16_t)(y)))
#define __builtin_divsd(x,y) (( int16_t)((( int32_t)(x))/(int16_t)(y)))
#endif

However tread very carefully here, as I recall having huge heartache at the
time with getting the nuances just right, and hence the leftover mess..

There's the macro solution (simple and consistent with everything), the
Microchip library implementation, and Bills assembler magic. Watch for
issues with the way the types are declared - for I believe the email
archives feature a discussion of at least on of these functions leading to
a rather nasty bug.

phollands:
In think the __builtins.h discussion is all a "red herring" (something that is not really an issue, leading us off in the wrong direction). I suspect the MPLAB-X dynamic in line parser, wrongly showed that the macros applied to our normal code, when they do not.

I now see that for the XC16 compiler and C30 compiler, in udbTypes.h, that the macros are only included for compilers that are not for the dspic. i.e. the source is
```
// Device header file
#if defined(__XC16__)
#include <xc.h>
#elif defined(__C30__)
#include "Compiler.h"
#else
#include "builtins.h"
#endif // __XC16__
```

builtins.h is not included anywhere else but in udbTypes.h with those defines above.

Sorry if I took us all in a wrong direction there for a while. But it was good to refresh. So all of my testing over the summer was done with the normal dspic hardware doing the fast divides and multiplies. I will check the assembler code as well in a moment.

And here is the disassembly of the compiled code as proof...

```
247:               	// When there is a chance that data will be read soon,
248:               	// have the new average values ready.
249:               	if (sample_count > ALMOST_ENOUGH_SAMPLES)
00F29C  80AF51     MOV sample_count, W1
00F29E  200D80     MOV #0xD8, W0
00F2A0  508F80     SUB W1, W0, [W15]
00F2A2  34003C     BRA LE, 0xF31C
250:               	{
251:               		udb_vcc.value = __builtin_divsd(udb_vcc.sum, sample_count);
00F2A4  80AF52     MOV sample_count, W2
00F2A6  80AEE4     MOV 0x15DC, W4
00F2A8  80AEF5     MOV 0x15DE, W5
00F2AA  980714     MOV W4, [W14+2]
00F2AC  980725     MOV W5, [W14+4]
00F2AE  90021E     MOV [W14+2], W4
00F2B0  9002AE     MOV [W14+4], W5
00F2B2  090011     REPEAT #0x11
00F2B4  D82A42     DIV.SD W4, W2
00F2B6  88AEC0     MOV W0, 0x15D8
```

The actual divide is done with a single instruction 
DIV.SD W4, W2

Best wishes, Pete

phollands:
My next move will be to check how the exception handling works. If the main program is re-started without turning off the existing interrupts (e.g. DMA interrupt), then sample_count will be set to zero during re-initialisation even while the DMA interrupt is working and possibly dividing using a denominator of sample_count (zero), which would then trigger a divide by zero exception. So I will try to look at that when I have time with the debugger, Pickit 3, and real hardware. In the past when I manually generated divide by zero traps, the microcontroller rebooted, and the telemetry was designed to record the reason for the trap / reboot, along with the offending address of the program counter at that time.

Inspirati:
ah-ha *headache ensues*

Jeffem:
Hi Pete,

I apologize if I use this canal to speak directly with you and not on the forum.

Be free to tell me if it is not appropriate!

 

As you know, I am trying to migrate my Quadcopter code in the current official version MP5.

 

I observe a very important improvement of the code between MP4 and MP5! Even more clear and struturate than the MP4 that I love to work in.

(But I have still difficulties with your awful use of callback;)

 

I send you this email because I have noted a change that I do not understand:

In background .c, the previous instruction was 

#if (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)…

And now, in MP5, it becomes 

#if (BOARD_TYPE != UDB4_BOARD && HEARTBEAT_HZ == 200)…

 

That means that my UDB5 need the MPU to run the T1 interrupt.

But the instruction in MPU6000.c is still the same :

#if (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)

                //  trigger synchronous processing of sensor data

                _T1IF = 1;              // trigger the heartbeat interrupt

#endif // (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)

 

So, theorically (I have not run it on the real board), this T1 interrupt routine will never be executed on MP5.

 

Do you agree with me?

 

Best regards, gfm

Inspirati:
which reminds me, the concept behind my implementation of setjmp/longjmp
was to avoid a total reset (which involves the pre-main() startup code
clearing most of memory, unless specifically customised not to). Some of my
beta branches would have featured alpha efforts to avoid such things as
calibration, gps initialisation (which would always fail horribly), startup
'waggles', and the like..

Inspirati:
G'day Jeffem,

I suspect what you a seeing is correct, however to confirm such would take
far more investigation than I'm able to tackle just now.

Please note, integrating quad support is something i worked on extensively
some years back, and is in fact the only airframe which I ever flew with.
Not that I recommend it, but you may find records of my work in the
MatrixPilot_beta branch..

phollands:
GFS, I think you copied in github when you emailed me. So here is a brief answer:-
For the UDB5, the entire heartbeat() is driven at 200HZ by the mpu6000 dma interrupt (priority 6) that transfers data into memory once the SPI 2 interrupt has occurred for receiving serial data from the MPU6000. There is no dedicated heartbeat() timer for the UDB5. The function heartbeat() is driven by the receipt of data from the MPU6000. heartbeat() then creates a software interrupt at priority 3, which in turn calls heartbeat_pulse() which runs at priority 3 (much lower), can calls the DCM algorithms for the IMU, before calling the flight_controller() routines to run the plane.  Best wishes, Pete

Jeffem:
Ok, Peter,

In that case, I must have, in the IMU read routine, 

#if (BOARD_TYPE == UDB5_BOARD && HEARTBEAT_HZ == 200)

                //  trigger synchronous processing of sensor data

                _T1IF = 1;              // trigger the heartbeat interrupt

#endif // (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)

 

And not the current (MP5) and old (MP4) test:

#if (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)

                //  trigger synchronous processing of sensor data

                _T1IF = 1;              // trigger the heartbeat interrupt

#endif // (BOARD_TYPE != UDB5_BOARD && HEARTBEAT_HZ == 200)

 

Or, to be coherent with the (new?) MP5 declaration for T1 interrupt configuration, in background.c|static inline void init_heartbeat(void) which is:

#if (BOARD_TYPE != UDB4_BOARD && HEARTBEAT_HZ == 200)

 

The same compiler if condition in the IMU read routine must be:

#if (BOARD_TYPE != UDB4_BOARD && HEARTBEAT_HZ == 200)

 

I insist on that logic because I configure the IMU sample rate at 1 KHz.
If we prefer to control the heartbeat rate with the IMU and not with the Timer 1, this fundamental clock become IMU sample rate dependant and so the configuration of the IMU is not innocent!

 

gfm

 

De : Pete Hollands [mailto:notifications@github.com] 
Envoyé : mercredi 23 novembre 2016 15:11
À : MatrixPilot/MatrixPilot
Cc : Jeffem; Author
Objet : Re: [MatrixPilot/MatrixPilot] Test of the divider sample_count before calling the __buitin_divsd (#98)

 

GFS, I think you copied in github when you emailed me. So here is a brief answer:-
For the UDB5, the entire heartbeat() is driven at 200HZ by the mpu6000 dma interrupt (priority 6) that transfers data into memory once the SPI 2 interrupt has occurred for receiving serial data from the MPU6000. There is no dedicated heartbeat() timer for the UDB5. The function heartbeat() is driven by the receipt of data from the MPU6000. heartbeat() then creates a software interrupt at priority 3, which in turn calls heartbeat_pulse() which runs at priority 3 (much lower), can calls the DCM algorithms for the IMU, before calling the flight_controller() routines to run the plane. Best wishes, Pete

Inspirati:
Please beware that any pre-processor checks on the UDB5 define needs to
include the AUAV3 as well - since the AUAV3 also features the MPU6000.

Perhaps we'd do well to have an independent define for the MPU6000 instead,
and define this inside the libUDB/Config<board-type>.h files.

Jeffem:
Hi Robert,

Thanks for this information.

When I look for the Multicopter in MP, I found the work of MarkW, in the branch “Multi”.

Unfortunately, this branch is dedicated to the UDB4 while I got an UDB5.

I plucked up my courage and I proceeded to migrate the Mark’s work in the UDB5 basis MP_4.1

In parallel I studied the Quad behavior with Matlab/Simulink.

The basis of my model is derived from Paul Bizard’s model.

I have a great admiration for the elegant and simple solution applied in the DCM principles.

So I decided to keep the Bill & Paul philosophy in my implementation:

·         Re-use as much as possible rmat,

·         Avoid complex operation like trigonometric or square root computations,

·         Keep the guidance and pilot structures (pitchcntrl, navigate,…).

More of that, I intend to command the quad like a plane and not as a copter: keeping “the ball in the middle” i.e.  coordinated flight, autopilot looking forward.

 

As I said in other messages, I am sticked today on the vibration treatment and I stopped software development.

I will come back in your MatrixPilot_Beta branch to pick up any good solutions!

 

Best regards,

gfm

phollands:
Guy, as the additional test of sample_count > 0 in your Pull Request (PR),  is not the correct solution improving a potential divide by zero operation, I am going to close this PR.

Jeffem:
Hi Pete,

I am interested by your request Interrupt priority analysis #122

 

I am sorry to come back on that subject in object : “sample count dividing by zero” but I was thinking about new hypothesis that I would like to expose to you.

I have upgrade the report that you know about my activities on MP.

I join an extract about the chapter in object.

 

I am happy to read your information about your request Interrupt priority analysis #122 because I intend to do the same thing for trying to understand this dividing by zero exception. 

Furthermore, as I have changed the interrupt logic in the Quad adaptation of MP, I must check the good behavior of these changes.

Your work and method will be very useful.

I have not yet got the RS232 analyzer, it is not my priority now, but it is in my todo list.

Best wishes,

gfm

 

 

De : Pete Hollands [mailto:notifications@github.com] 
Envoyé : mardi 6 décembre 2016 11:57
À : MatrixPilot/MatrixPilot
Cc : Jeffem; Author
Objet : Re: [MatrixPilot/MatrixPilot] Test of the divider sample_count before calling the __buitin_divsd (#98)

 

Guy, as the additional test of sample_count > 0 in your Pull Request (PR), is not the correct solution improving a potential divide by zero operation, I am going to close this PR.

PR #99, opened 2016-11-22
general housekeeping, cleanup whitespace, dead-code, executable flag …

Inspirati:
…under *nix builds, build system defines for LIN versus NIX, build script includes, conditional compilation faults with OSD, c/c++ include specs, and quiet warnings WRT magnetometer code being enabled

PR #100, opened 2016-11-22
Code Housekeeping Example with commits grouped together

MatrixPilot:
Robert,

I've taken your PR from #99 , and I have re-presented them here in this PR, to show how I think the commits can be grouped together to make the PR and the resulting commit history in master, much easier to read. 

If you think this PR is OK. Please can you approve it and merge, and then close PR #99 withoutmerging that one.

BTW: I think there is still some work to do on magnetometer options, but we can put that in another PR.

Best wishes, Pete


Inspirati:
Nice clean-up of my clean-up PR Peter, thanks (-:

PR #106, opened 2016-11-28
update build system to include new Configs and also change the target…

Inspirati:
… name to be prepended with the config rather than the compiler (aka, compiler now deprecated)


phollands:
Nice to see the test environment coming along.

PR #115, opened 2017-01-24
Covscan fixes

Inspirati:


phollands:
Robert, Thanks very much for these fixes. Merging. Best wishes, Pete

PR #118, opened 2017-02-02
Ublox: Update dynamics conf to use Airborne < 4g

phollands:
The configuration had been using Automotive which assumes primarily 2D navigation.
The maximum velocity rate for Z axis in earth is 15m/s with that setting (see table below).
I know that I have been close to reaching that when flying.

I would like to suggest that we move to using "airborne <4g".

Comments welcome as always. Best wishes, Pete

[The manual ](https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_%28UBX-13003221%29_Public.pdf?utm_source=en/images/downloads/Product_Docs/u-bloxM8_ReceiverDescriptionProtocolSpec_%28UBX-13003221%29_Public.pdf) for that dynamic configuration in the Ublox 8 says:-

**7.1 Platform settings**
u-blox receivers support different dynamic platform models (see table below) to adjust the navigation engine to the expected application environment. These platform settings can be changed dynamically without performing a power cycle or reset. The settings improve the receiver's interpretation of the measurements and thus provide a more accurate position output. Setting the receiver to an unsuitable platform model for the given application environment is likely to result in a loss of receiver performance and position accuracy.

**Dynamic Platform Models**

- Stationary Used in timing applications (antenna must be stationary) or other stationary applications.
- Velocity restricted to 0 m/s. Zero dynamics assumed.
- Pedestrian Applications with low acceleration and speed, e.g. how a pedestrian would move. Low
- acceleration assumed.
- Automotive Used for applications with equivalent dynamics to those of a passenger car. Low vertical
- acceleration assumed.
- At sea Recommended for applications at sea, with zero vertical velocity. Zero vertical velocity
- assumed. Sea level assumed.
- Airborne <1g Used for applications with a higher dynamic range and greater vertical acceleration than a
- passenger car. No 2D position fixes supported.
- Airborne <2g Recommended for typical airborne environments. No 2D position fixes supported.
- Airborne <4g Only recommended for extremely dynamic environments. No 2D position fixes supported.
- Wrist Only recommended for wrist worn applications. Receiver will filter out arm motion. (just available for protocol version > 17)

screen shot 2017-02-02 at 22 09 29


KeesGuijt:
Pete,

I agree that Airborne will probably perform better than Automotive.

Best wishes,
Kees
Ps: mind the "<"

phollands:
Thanks Kees. 

I also enclose below a small utility that I wrote for checking CRC's and then obtaining the new CRC after modifying the message. The end printout requires the bytes to be reversed when used. i.e. 0xff17 (shown below) becomes 0x17 0xFF in the message data structure.

```
Peters-iMac:checksum phollands$ cat checksum.c
// Program to calculate CRC checksum's for Ublox messages
// Pete Hollands, 2nd Feb 2017

#include <stdio.h>
#include <stdint.h>

uint16_t ubxCrc(const uint8_t* data, uint16_t) ;

const uint8_t config_NAV5[] = {
	//0xB5, 0x62, // Header
	0x06, 0x24, // ID
	0x24, 0x00, // Payload length
	0xFF, 0xFF, // Bit Mask, 0XFF means apply all of the config below
	0x08,       // Dynamic Model Number, Airborne with <4g Acceleration
	0x02,       // Position Fixing Mode. 3D only.
	0x00, 0x00, // Fixed altitude (mean sea level) for 2D fix mode only.
	0x00, 0x00, // Part of fixed altitude above (4 bytes in total)
	0x10, 0x27, // Fixed Altitude variance for 2D Mode (4 bytes)
	0x00, 0x00, //
	0x05, 0x00, // Minimum Elevation of Sats in degrees (1 byte). Dead Reckoning Limit in Seconds.
	0xFA, 0x00, // Position DOP Mask (2 bytes)
	0xFA, 0x00, // Time DOP Mask (2 bytes))
	0x64, 0x00, // Position Accuracy Mask (2 bytes) Meters)
	0x2C, 0x01, // Time Accuracy Mask (2 bytes))
	0x00, 0x00, // Static Hold Threshold (1 bytes). DGPS timeout (1 bytes))
	0x00, 0x00, // Reserved
	0x00, 0x00, // Reserved
	0x00, 0x00, // Reserved
	0x00, 0x00, // Reserved
	0x00, 0x00, // Reserved
	0x00, 0x00, // Reserved
	//0x13, 0x77  // Checksum
};

int main(void)
{
	uint16_t i  = 0;
	i = ubxCrc(config_NAV5,40);
	printf("The crc is 0x%04hx\n", i);
}
uint16_t ubxCrc(const unsigned char* data, uint16_t size) {
    uint16_t crc_a = 0;
    uint16_t crc_b = 0;
    if (size > 0) {
        do {
            crc_a += *data++;
            crc_b += crc_a;
        } while (--size);
        crc_a &= 0xff;
        crc_b &= 0xff;
    }
    return (uint16_t)(crc_a | (crc_b << 8));
}
Peters-iMac:checksum phollands$ ./a.out
The crc is 0xff17
```

PR #119, opened 2017-02-03
mathlibNAV.c: Improve rounding in rotate_2D()

phollands:
An improvement to rotate_2D(). The fix ensures that rounding is effectively done between -0.49 to +0.5 rather than from 0,0 to 1.0. So this fix can remove a small bias in some calculations.


phollands:
Will effect the routines shown in the call graph below ...
screen shot 2017-02-04 at 08 05 48

PR #123, opened 2017-02-10
Proj gen fix

Inspirati:
Invert the logic of the make conditional so the default build appears first. Project generator does not interpret the conditional and therefore I broke it when I added support for the PX4 target.


phollands:
Thanks Rob.

Issue #124, opened 2017-02-17
Barometers.c returns wrong pressure with Test Dataset for BMP180 / BMP085


phollands:
If I define TEST_WITH_DATASHEET_VALUES in barometer.c in master,
and compare the result in the telemetry with the expected [answer from the data sheet](https://cdn-shop.adafruit.com/datasheets/BST-BMP180-DS000-09.pdf) [See page 15], the resulting answer for the pressure is wrong. The relevant portion of telemetry from the F2: format line of SERIAL_UDB_EXTRA is:-

:tmp150:**prs7831**:alt0:

The temperature should show 150 (15 degrees), which it does.
The pressure should show 69964 but instead is showing 7831.

I had previously done a test flight using the barometer [Adafruit BMP180 ](https://www.adafruit.com/product/391) on a UDB5 providing the altitude. The altitude hold was good. But when viewed later on Google Earth, the Altitude of the GPS route when compared with the altitude of the IMU (based on the barometer) differed by more and more, as the plane flew higher and higher. So I then started to question whether the calculations in the barometer were correct. They were consistent, but they do not look correct.  

This is the first time I have used the barometer code in MatrixPilot.
I shall have a poke around and see if I can spot something obvious.

Best wishes, Pete

P.S. I enclose a screenshot of the KMZ GoogleEarth view showing the GPS track as much higher than the IMU track when the plane is high. But lower down, you will see that the two track re-converge and get closer together when the plane is close to the origin altitude at ground level.

![screen shot 2017-02-17 at 10 16 02](https://cloud.githubusercontent.com/assets/542066/23061566/57c5a470-f4fa-11e6-8e32-0e49b1039419.png)

elgarbe:
Peter, [here](https://github.com/MatrixPilot/MatrixPilot/blob/master/libUDB/barometer.c#L206) b7 is defined as LONG, but datasheet say that it need to be unsigned long. Could it be a source of error?

elgarbe:
Ac4, 5 and 6 need to be unsigned short and MP code define they as unsigned short (int for our 16 bit uC) [here](https://github.com/MatrixPilot/MatrixPilot/blob/master/libUDB/barometer.c#L43) so that is ok

KeesGuijt:
Pete, Leo,

The pressure value appears to be a bug. 
[...]

PR #125, opened 2017-02-18
Fix barometer pressure

phollands:
This is a fix for issue #124 .

### DIFF
--- a/libUDB/barometer.c
+++ b/libUDB/barometer.c
@@ -37,35 +37,35 @@
 typedef union
 {
 	struct {
-		int ac1;
-		int ac2;
-		int ac3;
-		unsigned int ac4;
-		unsigned int ac5;
-		unsigned int ac6;
-		int b1;
-		int b2;
-		int mb;
-		int mc;
-		int md;
+		int16_t ac1;
+		int16_t ac2;
+		int16_t ac3;
+		uint16_t ac4;
+		uint16_t ac5;
+		uint16_t ac6;
+		int16_t  b1;
+		int16_t  b2;
+		int16_t  mb;
+		int16_t  mc;
+		int16_t  md;
 	};
-	unsigned char buf[22];
+	uint8_t buf[22];
 } barCalib_union_t;
 
 static barCalib_union_t bc; // barometer calibration constants
-static long b5; // this is the important number for calculating the temperature
+static int32_t b5; // this is the important number for calculating the temperature
 
-static const unsigned char bmp085read_barCalib[] = { 0xAA };  // Address of the first register to read
-static const unsigned char bmp085read_barData[]  = { 0xF6 };
+static const uint8_t bmp085read_barCalib[] = { 0xAA };  // Address of the first register to read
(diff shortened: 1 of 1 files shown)


phollands:
Flight Test on UDB5 went OK.  The GPS Altitude Track and Dead Reckoning Track (from barometer) are close to each other in normal flight and climbs. The screenshot below shows both the IMU Track and the GPS track together.

 Pressure value in SERIAL_UDB_EXTRA shown as:- 
:**tmp208:prs101502:alt0**:bmv0:mA10:mAh3:DH0: which is correct for my location and height at origin.

screen shot 2017-02-18 at 17 07 50

So if there are no further comments suggesting changes, I shall merge this in tomorrow.

phollands:
Ground Tested the AUAV3 with barometer. Tests OK. 
(slightly higher temperature, but tested inside, and is situated on main AUV3 board.)
**tmp262:prs101381**:alt0:

Best wishes, Pete

elgarbe:
Peter, do you plan to merge the code today? I'm preparing a mission today (I want to test my 1.2GHz biquad antenna) and could hook up my barometer to try it.
best wishes!

phollands:
Hi Leo,

Your email crossed as I merged the code at the same time. All done.

Best wishes, Pete

phollands:
One thing that I have been thinking about. Does the barometer code work
with the magnetometer code at the same time on the same I2C interface ? As
I am one of the few people that use the magnetometer successfully, I
suspect that this currently is only an issue for myself. I will get to
testing that soon.

Best wishes, Pete

elgarbe:
I have one of this: http://selfbuilt.net/shop/gy-80-inertial-management-unit
so in any time I could test magnetometer as well. But I'm concern about calibration, I think that I read about the difficult to calibrate this sensor.

Inspirati:
the mag and baro 'should' work on the same bus together - albeit it's been
a long time since i wrote that code, and it was around the same time as i
wrote the baro code which it seems you've just found to be in error, so you
mileage will no doubt vary. for memory, the code which handles the two
devices together is somewhat convoluted as they work at different sampling
rates, and one requires several more transactions to get a result than the
other. thus i suspect it may all look somewhat confusing (at least at first)

phollands:
Thanks Rob, I'll take a look before I fly.

phollands:
Leo, Yes, I have **not** yet been successful with the HMC5883L, but I do have success with the older HMC5843 chip.  The offsets on the HMC5883L that I have in my possession  are much larger than the 5843, and I think that Bill's automatic calibration of offsets for the 5883L is too slow. I looked into it a couple of years ago, and made some tools for visualising the offsets in 3D (using Python). It could be fixed if we were motivated to do that (i.e. there was a good reason. e.g. flying without GPS).

Hope you have managed to get a good flight in.

Best wishes, Pete

elgarbe:
I have a really good flight today. I will post it later. 10km away and back. Barometer seems to work excelent.

phollands:
Leo, That is good news. Look forward to seeing your post on uavdevboard.

PR #126, opened 2017-02-21
I2C2 improvements to use mag and barometer together

phollands:
The goal is to allow the Barometer and Magnetometer to reliably work together while sharing the same I2C2 bus (2 wires).

Once calibrated the Magnetometer state machine can return new data from the sensor with one call. But the Barometer state machine expects 6 calls to return the temperature and pressure. We intend to retrieve useful sensor data from each device bat 4Hz. 

It is not possible to just toggle calls between each device, as sometimes, it is essential that one device is called up to 6 times consecutively, for each of the next 40Hz frames. Rob had a good go at this, but I have really taken his ideas further.  I have modified the rxBarometer() and rxMagnetometer() routines so that they return a state as to whether they needs servicing imminently again. If they do, then they will be called again on the next 1/40th of a second pulse. However, the barometer is limited in doing this for a maximum of 6 repeated calls, and the magnetometer can repeat for a maximum of 4 times, which fit's it's own needs on the I2C bus. 

I am posting this Pull Request early for comments, so that I don't go too far down the route of testing, if I then need to make significant changes to this. Testing may take a while; I shall probably needs to now get the latest Salae Logic Analyser which incorporates a form of oscilloscope as well as logic analysis; When joining several I2C devices together with wiring, it will be important to check that the capacitance of the wires, along with the pull resistors are all working correctly.  This will not be so important when all the devices are on one board. e.g. on the AUAV3, or the board that Leo has found.

I note that we run our I2C2 devices at 200K Hz which is an unusual speed. The spec is usually for 100K Hz, and then there is a variant specification for 400K Hz. I have left the speed at 200K as it has been working for us so far.

Best wishes, Pete

### DIFF
--- a/libDCM/libDCM.c
+++ b/libDCM/libDCM.c
@@ -82,59 +82,68 @@ static boolean gps_run_init_step(uint16_t count)
 	return false;
 }
 
-#if (USE_BAROMETER_ALTITUDE == 1)
 
-// We want to be reading both the magnetometer and the barometer at 4Hz
-// The magnetometer driver returns a new result via the callback on each call
-// The barometer driver needs to be called several times to get a single 
-//  result set via the callback. Also on first invocation the barometer driver
-//  reads calibration data, and hence requires one extra call
+// We want to be returning reading from both the magnetometer and the barometer at 4Hz.
+// However each device may need to be called many times before a reading is returned.
+// The code below ensures that a given sensor is called repeatedly until it reaches
+// a state where the other device can be called safely over the same I2C2 bus, even if 
+// the first device has not yet returned a complete set of sensor data.
+// For example, at startup the barometer needs 12 calls before returning pressure.
+// But in normal operation it will need 6 calls.
+// By contrast in normal operations the magnetometer only needs one call, once it is calibrated.
 
-void get_data_from_I2C_sensors(void)
+void get_data_from_I2C_sensors(void) // Expected to be called at 40Hz
 {
-	static int toggle = 0;
-	static int counter = 0;
-
-	if (toggle) {
-		if (counter++ > 0) {
-//#if (MAG_YAW_DRIFT == 1 && HILSIM != 1)
-#if (MAG_YAW_DRIFT == 1)
-//			printf("rxMag %u\r\n", udb_heartbeat_counter);
-			rxMagnetometer(mag_drift_callback);
-#endif
-			counter = 0;
-			toggle = 0;
-		}
-	} else {
-		rxBarometer(udb_barometer_callback);
-		if (counter++ > 6) {
-			counter = 0;
-			toggle = 1;
+	static uint8_t barometer_needs_servicing = BAROMETER_SERVICE_CAN_PAUSE;
+	static uint8_t magnetometer_needs_servicing = MAGNETOMETER_SERVICE_CAN_PAUSE;
+	static uint8_t counter_40Hz = 0;
+	
+	// Split a 1 second interval (40 calls) into 8 interleaved segments using the following logic
+	if ((counter_40Hz == 0) || (counter_40Hz == 10) || (counter_40Hz == 20) || ( counter_40Hz == 30))
+	{
+#if (USE_BAROMETER_ALTITUDE == 1 && HILSIM != 1)
+		barometer_needs_servicing = rxBarometer(udb_barometer_callback);
+#endif // (USE_BAROMETER_ALTITUDE == 1 && HILSIM != 1)
+	}
+	else if // Allow a total of 6 consecutive calls to the barometer to return pressure
+		(((counter_40Hz >= 1) && (counter_40Hz <= 5))
+		|| ((counter_40Hz >= 11) && (counter_40Hz <= 15))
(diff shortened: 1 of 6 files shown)


phollands:
The BMP180 only needs 4.5ms to gather the temperature ([see page 11 of datasheet PDF](https://cdn-shop.adafruit.com/datasheets/BST-BMP180-DS000-09.pdf)), so I have removed the pause of 50ms introduced by (barCalibPause = 2). As each call is 25 milliseconds apart, that will suffice. But in the case of the Pressure, it needs 25.5 ms when oversampling is set to 3 (which it is), and so I have reduced it's delay to barCalibPause = 1. So A pause that will give it 50 ms to complete the 3 pressure samples, before the combined total is read back to MatrixPilot. The logic analyser shows the traffic to all be correct (barometer only, no magnetometer yet), with no-retries on my system which has about 7 inches of cabling between the UDB5 and Barometer.

phollands:
The  internal barometer and internal magnetometer are working for the AUAV3 in ground tests, in that sensible magnetometer and barometer values are returned in the telemetry. So I believe the code to be working correctly.

However, I cannot see the actual I2C2 bus on the AUAV3 as I don't have access to those very small pins. I will conduct a test flight with the AUAV3 after I've done a manual calibration of soft and hard iron offsets for the HMC5883 in the AUAV3 mag. That magnetometer can have very large offsets which are not handled by Bill's automatic in flight calibration code, but I'm hoping that [using a manual calibration may fix that.](http://theboredengineers.com/tag/hmc5883l/)

For my UDB5, I have also ordered a Sparkfun BMP180 where I can  de-select the use of the resistors to hold the SDA and SCLK lines high (as that is already done by the magnetometer). My existing Adafruit BMP180 has no facility of de-connecting the resistors.  In contrast to the AUAV3, with these external devices, I should then be able to use the logic analyser to confirm that all the traffic I2C traffic is proceeding smoothly when both barometer and magnetometer are operating.

phollands:
Flight Tested (00125 Green Tip Alpha Axion and AUAV3), with PR #128 also applied (to have good magnetometer offsets). Also extensive testing with SERIAL_MAGNETOMETER to show that the state machines of rxMagnetometer and rxBarometer were running correctly.  Barometer and Magnetometer working acceptably during flight. I could still improve the enclosed picture with more careful manual calibration of the static offsets.

screen shot 2017-02-28 at 12 54 44

elgarbe:
Peter, great work!
So, hcm5883 could work ok after a good offset calibration?

phollands:
Leo, Yes your combined barometer and magnetometer (HMC5883)  board will work at a detailed sensor acquisition level. You just need to be sure that you do not have any problems associated with the correct resistor termination and the capacitance of the I2C wires that connect back to your main autopilot board.

**I note that you are in the Southern Hemisphere, so your magnetism may be reversed. I'm not sure if the code takes that into account anywhere at all ! So please check that carefully**.

Also you will need to get the orientations correct of your magnetometer chip, and calculate the magnetometer offsets carefully. Bill built in a feature called "auto-alignment" that should align the RMAT DCM matrix axis with the magnetometer matrix if they are by a few degrees. In fact even if the magnetometer axis is way out of alignment with the axis of the gyros / DCM, it is still meant to work. I have not tested that. The following diagram may be useful for you ...




If you want to use this feature, I would advise:
Set it up in the plane, and do the calibration of offsets with SERIAL_MAG_CALIBRATE. Check the body magnetometer readings carefully, and check that they all look good in scatter plots of X and Y, X and Z, and Y and Z. When you have the offsets correct, you should have nearly perfect circles centred on the origin of 0,0. The SerIAL_MAG_CALIBRATE telemetry file is designed to be used as a CSV file for that purpose. 

When you do the calibration, try to do the turns of your plane in the axis of Magnetic South to Magnetic North. You are trying to capture the maximum and minimum (including negative values) of the earth's field strength, in the axis of the body magnetometer (X,Y,Z) axis.

Because you will have I2C wiring (which may increase capacitance of the signals held high by the resistors), after calibration, I would run the SERIAL_MAGNETOMETER debug telemetry. This should show the magMessage  increasing once per telemetry sample, and the message number should be 7 each time. If it is anything else, then you are having trouble on the wires. When you run this test, again, yaw the plane around slowly over a few second with the wings level. Do that again. Then yaw the plane around with the plane rolled at 45 degrees. Do that again. Then with the plane rolled over on it's side at an angle of 90 degrees. Do that again. Afterwards, in the telemetry, check that the Earth Magnetometer readings are always pointing North. i.e. Y should be positive and large, while X should very small. You can see that easily if you do a scatter plot of the mag earth X  and mag earth Y readings. 

Best wishes, Pete

phollands:
Leo, do you have the HMC5883 or the HMC5883L ?

elgarbe:
Thank Peter!
I will try it in the next weeks!
One last question, I would like to use UDB5 on my father sailplane (3 ch, rudder, throtle and elevator) but don't want to use GPS. Just stabilization and altitud control. What feature add the magnetometer in this case?

phollands:
Hi leo, Let's take this discussion out of github and this PR now. I'll
email you one on one. Best wishes, Pete

PR #128, opened 2017-02-26
Static mag offsets2

phollands:
In order to test the AUAV3 with both barometer and magnetometer running on the I2C bus, I needed to get the HMC5883L running correctly.  I did this eventually by setting up static offsets, and switching off the automatic calibration of offsets in the magnetometer. My hypothesis is that the HMC5883L has larger offsets than the original HMC5843 which Bill used when developing automatic calibration, and so it is either too slow or does not work on the HMC5883L. I tidied up some of the code for Static Offsets, and have added a new printout feature (SERIAL_MAG_CALIBRATE) for measuring the offsets in the telemetry. After carefully running that facility, and creating graphs in the corresponding csv file, I now have a pretty good round and centred sphere of measurements from the HMC5883L.  Ground testing in the garden, and plots in Google Earth, show this to be working much better now. The earth rotated magnetic vector is probably still moving at between +/- 10 degrees depending on roll and pitch (including inverted flight), but that is OK for now. (I think it could be improved still further with careful calibrations).

So I intend to put rebase base this set of  commits onto the previous set of commits #126 , and run a flight test when the weather allows using the AUAV3 barometer BMP180 and HMC5883L magnetometer together.

For reference my AUAV3 HMC5883L magnetometer offsets were:-
```
#define MAG_STATIC_OFFSET_X       70
#define MAG_STATIC_OFFSET_Y      132
#define MAG_STATIC_OFFSET_Z       -3
```

I show below the plots of measurements of turning my test board around in circles in each axis:-

screen shot 2017-02-26 at 23 05 39

Best wishes, Pete

### DIFF
--- a/Config/options.h
+++ b/Config/options.h
@@ -358,6 +358,7 @@
 // SERIAL_CAM_TRACK is used to output location data to a 2nd UDB, which will target its camera at this plane.
 // SERIAL_MAVLINK is a bi-directional binary format for use with QgroundControl, HKGCS or MAVProxy (Ground Control Stations.)
 // SERIAL_MAGNETOMETER outputs the automatically calculated offsets and raw magnetometer data.
+// SERIAL_MAG_CALIBRATE is used to calculate  magnetometer offsets for a static non changing calibration. 
 // Note that SERIAL_MAVLINK defaults to using a baud rate of 57600 baud (other formats default to 19200)
 
 #ifndef SERIAL_OUTPUT_FORMAT
--- a/Config/options_magnetometer.h
+++ b/Config/options_magnetometer.h
@@ -62,7 +62,17 @@
 //#define MAG_FLIPPED
 //#define MAG_DIRECT
 
+// Uncomment the following line for using Static Magnetometer Offsets. 
+// You will need to manually calibrate your plane's offsets in advance.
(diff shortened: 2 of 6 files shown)


phollands:
I have successfully tested the magnetometer in the AUAV3 (HMC5883) with static offsets, in flight.
See #126 for a picture from the GoogleEarth view of the magnetometer rotated by RMAT into the earth reference.

Some points that came up during testing with barometer and magnetometer together: 

- Sunlight did hit the barometer briefly during ground testing this morning. It made the IMUlocationZ jump by more than 100 meters. That is partly due to the effects of sunlight on a barometer (note there was not hear in this winter sun, just light)

screen shot 2017-02-28 at 13 29 20

- It is also partly also due to the fact that GPS latency compensation code is being applied to the barometer, when it should not. That is a separate issue to resolve beyond this PR. 
- If you reverse the AUAV board so that it's orientation is backwards, the you do need to set MAG_BACKWARDS in magnetometer.h. If you have already calibrated your MAG_STATIC_OFFSET_X, Y and Z before reversing, then you will need to do a sign change of X and Y offsets. Most people should do the calibration of the offsets with the board setup in the correct orientation in the plane, in which case this last issue will not happen; Your calculated offsets will be correct for your plane's setup.

The HMC5883 magnetometer is working with manual offsets with this PR, and also works at the same time as the barometer on the same bus (after PR #126 ).

PR #131, opened 2017-03-02
Fix build all sh script to return "pass" or "fail" of session to Travis 

phollands:
At the moment, if the script build-all.sh has an error in the make of one of it's many configuration scenarios, it will report the error in the log, but Travis will still believe that the entire session ran smoothly.

I have edited the script, so that any one single failure will result in a Fail being returned to the caller of build-all.sh (Travis). I have tested it outside of Travis.

At the moment, the SIL configuration and one of the other configuration are commented out, as they are failing anyway. I will look at how to fix the up, and then put in a separate Pull Requests for those fixes. 

I did look at make a subroutine called "try()" which would keep track of whether a given command line had been successful. However I ran into the difficulties of borne shell evaluation of quotes with the DEFS='xxx yyy'  syntax being passed all the way to the Make subsystem correctly. I spent half a day on that, before returning to live with multiple if statements through out the code for now.

### DIFF
--- a/build-all.sh
+++ b/build-all.sh
@@ -1,68 +1,122 @@
 #!/bin/bash
+command_failed=false
 
 git config --get remote.origin.url && git symbolic-ref --short HEAD
 
 mkdir -p _build
 cd _build
-
+ 
 mkdir -p XPlane; cd XPlane
-make -j 8 -f ../../makefile TARGET_NAME=XPlane DEVICE=HIL
+#if ! make -j 8 -f ../../makefile TARGET_NAME=XPlane DEVICE=HIL
+#	then
+#	command_failed=true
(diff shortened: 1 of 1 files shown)


phollands:
That is Good. The Travis build has failed on current master, and it should return fail. From that point of view the script is working.  I will merge this in and then alter PR #130 (add static mag offsets to RollPitchYaw and other projects) and see whether the test build then completes with this new script, and Passes.

screen shot 2017-03-02 at 17 35 48

phollands:
OK. The Mag offset PR #130 has now completed it's build with this new build-all.sh script, and Travis now shows it as passing. So we've had a pass and a fail, and so the script is working.

PR #132, opened 2017-03-04
refactor gplane[] into two: gplane[] and gravity_vector_plane[]

phollands:
gplane[] has had two meaning in the code, depending on the timing of when you view it in the DCM thread of activity. The 1st meaning was to be the  (gravity - acceleration) in the body frame (the plane's frame). This 1st meaning was used as the basis of measurement for DeadReckoning. Later in the code, the gplane variable is re-used for a second meaning,  to find the gravity vector. This made for some confusing reading for new users to the code. So I've refactored the 2nd meaning have a new variable called gravity_vector_plane[].


phollands:
Tested with RollPitchYaw on UDB5.

PR #133, opened 2017-03-04
Fix flan.pyw waypoints generation when flight has Return To Landing segments

phollands:
This is a fix for flight analyzer to be able to process Leo's Flight 77 correctly, in which he flew towards waypoint 0, then entered Return to Landing and the RTL waypoint 0. This combination of events exposed a bug, and actually the fact that RTL waypoints were historically not originally in flan.pyw, so it really needed additional logic which is attached in the enclosed commit.

This fix also tries to ensure that he colour coding of RTL flight segments are different from the main autonomous flight plan.

### DIFF
--- a/Tools/flight_analyzer/flan.pyw
+++ b/Tools/flight_analyzer/flan.pyw
@@ -1052,8 +1052,12 @@ def write_placemark_preamble_auto(open_waypoint,current_waypoint,filename,log_bo
     print >> filename, """        <description>waypoint""",
     print >> filename, current_waypoint,
     print >> filename, "</description>"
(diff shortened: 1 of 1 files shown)


phollands:
Just to confirm that this version of master was fully flight tested today, and then correctly analysed in flight analyzer (flan.pyw). The AUAV3 ran the magnetometer and barometer together and was in command. The UDB4 and UDB5 followed through on the controls and only used the GPS (no magnetometer and no barometer). I  used "Return to Landing" and Absolute Waypoints several times in the flight, and the all plotted correctly afterwards. 

return_to_landing_allowing_for_wind

PR #134, opened 2017-03-13
Improve barometer pressure and altitude update rates

phollands:
In looking through recent logs I noticed that the pressure was only updating at 2Hz rather than 4Hz. This is the fix. The barometer state machine needs to be called 5 times consecutively to deliver on reading of pressure. I enclose a small sample from today's flight with this fix showing the pressure updating at 4Hz.

screen shot 2017-03-13 at 09 09 53

As well as this fix, I'm going to look at the altitude calculation logic.  On my plane, this is only returning one new altitude per second. So expect another 2nd commit to this PR soon. I would like the altitude to update at the GPS_RATE (2Hz in my case) rather than 1Hz.


phollands:
Barometer altitude is now calculated at the GPS rate. (In my case that is 2Hz). I enclose a screenshot of altitude data from a test flight of the latest commit.

screen shot 2017-03-13 at 12 13 10

KeesGuijt:
Pete,

The GPS_RATE check was to fix a bug that made barometer erratic at 4Hz. Maybe your recent work has fixed this. So please verify 4Hz works as well.

Kees

phollands:
Kees, 

Thanks for the info on the potential bug / problem. The rate of calculating temperature and pressure is independent of the original GPS_RATE in  estAltitude.c.  estAltitude.c  takes the existing temperature and pressure (which  is returned at 4Hz in my code), and then calculates the altitude using the low priority thread of GPS navigation at priority 2 at the GPS_RATE. I note that it is using ordinary floating point calculations which we normally avoid in MatrixPilot as they are expensive. It is also using the pow() function in floating point which will also be expensive. But then the GPS common navigation thread is running at a low priority, so other threads should not be effected. 

I  know that the barometer pressure and temperature seems to be updating correctly each time at 4Hz in my last two flights. But I will double check.

I will review the altitude estimation code, and the barometer code,  as to how it was before my changes, and then do some timing measurements on the new code with  the logic analyser to make sure all the key interrupts / processes are running at the correct time, and that none of them are taking too long.

Best wishes, Pete

phollands:
Kees,

I have done a flight test today, with the Ublox GPS running at 4Hz. I flew many climbs and descents, with relatively sharp changes of pitch, so that I could then compare the GPS altitude with the Barometer Altitude. The screenshot below, shows both the GPS track and the IMU Track (Barometer altitude). They were remarkably close to one another throughout. The GPS track is slightly smoother on altitude as it is clearly filtered using the velocity vector. But the Barometer is still remarkably good, and as potentially being more reliable than the GPS for altitude, although not today. Sattelites in view started at 6 on the ground, but moved through to 10 before take off, and remained at 10 or 9 for the entire flight. HDOP was at 4 for the entire log.

screen shot 2017-03-16 at 10 34 00

screen shot 2017-03-16 at 10 49 03

One point that I have looked at carefully on the graphs and data, is whether I can see any material effect of the fact that the barometer altitude is being treated as if it was from the GPS, and is having the same GPS latency algorithm applied to the barometer altitude as the GPS altitude. (Adding on previous differences to altitude). I could not discern any material difference in this data.

Looking through the old code, before this PR, it is clear that the barometer state machine needed 8 consecutive calls to return a reading. But the code was only making 6 consecutive calls. This would only have potentially become an issue if the magnetometer was being used at the same time. So I'm not sure Kees, if you were using the magnetometer with the barometer ? Anyway, as far as I can see, it's all working smoothly now with altitude being calculated at 4Hz from the barometer. Maximum CPU usage is 19%.

Unless their are further comments or suggestions, I'm minded to merge this PR.

Best wishes, Pete

KeesGuijt:
Pete,

Nice. I have similar results (but still at1Hz refresh rate). I see magnetometer is enabled as well, no conflicts apparently? 

Best wishes,
Kees

phollands:
Kees,

Yes, the magnetometer worked correctly for the whole flight when viewed in
the KMZ file with the magnetometer vector (rotated into Earth reference by
RMAT)  displayed.

The barometer and magnetometer now always with the same timing relative to
40Hz frames, regardless of whether the other sensor is enabled or not. That
should take out some of the uncertainty about how well they work with or
without the second sensor.

Best wishes, Pete

phollands:
Magnetic vectors from the same flight:-
screen shot 2017-03-16 at 15 18 54

PR #141, opened 2017-05-29
Reduce CPU Loading by revision of A/D for UDB5. 

wjpremerlani:
use Mark Whitehorn's timing constants to reduce CPU loading.

### DIFF
--- a/libUDB/analog2digital_udb5.c
+++ b/libUDB/analog2digital_udb5.c
@@ -46,7 +46,8 @@ uint8_t DmaBuffer = 0;
 uint16_t maxstack = 0;
 #endif
 
-#if 1   // these are the original/legacy values, switch this off to test Mark's new timing constants
+#if 0   // these are the original/legacy values, switch this off to test Mark's new timing constants
+// May-29-2017, WJP: Marks timing constants work well and they reduce CPU loading, so lets use them.
 #define ALMOST_ENOUGH_SAMPLES   216 // there are 222 or 223 samples in a sum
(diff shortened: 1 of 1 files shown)


MatrixPilot:
hey William, log into skype

phollands:
I am going to enclose Bill's comments on this PR which he has sent to me ....

Peter,

It turns out that Mark Whitehorn had already revised the sampling rate for the UDB5 A/D. However, his revisions were turned off. All I had to do was to turn them on.

The sampling rate is still a bit higher than it needs to be, but it is good enough, so in the interest of efficiency, lets go with Mark's values for now, we can always fine tune later.

Here are the CPU loading numbers:

legacy sampling rate, CPU loading at 24%
Mark's sampling rate at 16 MIPs, CPU loading at 13%
Mark's sampling rate at 32 MIPs, CPU loading at 6%.

At Mark's sampling rates, noise level is low and bandwidth is high. So, its a good engineering compromise between performance and CPU load.

Best regards,
Bill

MatrixPilot:
howdy guys, for memory this seems about right with me too. Marks
engineering of our equations is as sound as we might otherwise get.

phollands:
Thanks to Leo, in Argentina, who has kindly done a test flight (LOG0079,TXT)  of master to re-confirm that this change (reducing the A/D measurement rate on the UDB5) works fine for the calculation of battery voltage, amperage, and total current used.  The flight went smoothly. The CPU usage was reduced from 32% to 22% for his setup which uses On Screen Dispaly (OSD). I have reviewed the logs and all looks fine.

Issue #108, opened 2016-12-29
SERIAL_UDB_EXTRA buffer indexing bug at high baud rates (e.g. 921600 ).


phollands:
While experimenting with Baud Rates of about 1 Mega Baud ( 921600 baud rate in options.h ) linking a UDB5 to a Raspberry Pi A+, I started to get repeating snippets and slightly garbled lines of F2: telemetry. 

Using a logic analyser I was able to show that I was not dropping characters but that the issue was occurring within the UDB5 and MatrixPilot. 

My current hypothesis is that while vsnprintf was readying and copying data to the serial buffer, the interrupt routine, which is now churning through it's characters at a high rate and at priority 4, and eating up the buffer, reaches the end of text \0 null character; but that the vsnprintf, running at priority 3, has by then overwritten the \0 null character marking the end of text, and is partway through copying it's new vsnprintf output to the buffer but has not yet finished. vsnprintf (priority 3) can be interrupted by the transmission part of the serial code (priority 4).  The result it that the firmware prints past the original end of text and finds old previously printed text to re-send until finding a second null char further down the buffer. I should stress that is a hypothesis and not absolutely proven. 

I have never noticed this issue before, but I think it arises because the interrupt routine is now so fast at 1 Megabaud, that it is reaching the end of line null character quicker, and while vsnprintf is still preparing  and copying over the next message.

I have already coded up one test solution that works by stopping the transmission of characters while vsnprintf is working and copying data into the serial buffer. That solution does fix the issue, and I'll post it as a Pull Request once I've thought about this some more, and done some more testing.

phollands:
Fixed in PR #135 .

MatrixPilot:
another option that comes to mind would to the modify the uart tx interrupt
handler to replace each transmitted character in the buffer with a '\0'.

PR #142, opened 2017-06-04
Prevent the serial port from dropping a character when an extra TX in…

wjpremerlani:
…terrupt is generated when the port is stopped/started

### DIFF
--- a/libUDB/serialIO.c
+++ b/libUDB/serialIO.c
@@ -235,22 +235,26 @@ void udb_serial_stop_sending_data(void)
 void __attribute__((__interrupt__, __no_auto_psv__)) _U2TXInterrupt(void)
 {
 	_U2TXIF = 0; // clear the interrupt
-	indicate_loading_inter;
-	set_ipl_on_output_pin;
-	interrupt_save_set_corcon;
+    // prevent losing a character when stop and start sending data
+    if (U2STAbits.UTXBF == 0)
+    {
+        indicate_loading_inter;
+        set_ipl_on_output_pin;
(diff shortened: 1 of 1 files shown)


MatrixPilot:
When Bill started using a clock frequency of 32 megahertz, the SUE telemetry was dropping a single letter in the F13:  origin information message.

`F13:wek15761:origN472588539:origE113312005:origA58516:`

The word "week" had become "wek". After some joint debugging and analysis, we confirmed that when re-starting the serial ISR with a software interrupt, that if it is actually still writing writing a character out from it's shift register, and still has one to send also in the transmit buffer, then the new software interrupt, will cause a overwrite, with the new character, of the existing character in the transmit buffer. And so one character is lost.

I had previously introduced the ability for the serial IO to be stopped and then re-started each time the vsnprintf() routine is called to overcome another issue #108 with previous PR #135 . However,  if vsnprintf() runs and completes quickly, so that the re-start of serial IO  happens quickly before transmissions have actually stopped in the ISR (which is what was happening at 32 megahertz as opposed to 16 megahertz),  then we were losing a character. The F13 protocol message is one of the shorter of the SUE messages and simple for printf to process , as is the following message F20: . So vsnprintf completes quickly, the software restart of serial is initiated, but the serial hardware logic is still finishing off transmissions after having being told to stop, which is why  the bug first appeared in the F13: message.

The bug has not been seen at 16 megahertz, only at a 32 megahertz clock rate.

Thanks to Bill for his help in diagnosing the problem, and also for creating the fix.

PR #143, opened 2017-06-23
Fix brightness problems in video caused by overly bright white pixels…

tripoverzero:
… written by the OSD.

This fix drops the brightness of the white pixels to 80% of full-white. It fixes video that turns dark whenever the OSD is enabled.


phollands:
Very glad to see your contribution. Best wishes, Pete

PR #145, opened 2017-07-03
add 2nd battery voltage measurement and OSD data

elgarbe:
I test the code and works ok:
https://youtu.be/QMdhCW7jiAQ

In minim_osd.c I've created a second battery_update function. Not sure if it is ok or could add my code into current battery_update function.

BW, Leo


phollands:
Leo, Thanks very much for adding the 2nd Voltage for OSD users. Best wishes, Pete

PR #149, opened 2017-08-23
Keep track of GNSS parse errors in the telemetry; Flan  warns users if errors are present

phollands:
Bill has added code to the GNSS parsers for the EM406/506 and Ublox, to check if there are problems parsing the serial stream from the receiver units.  This then showed that The Ublox parser was having 8 errors in normal operations at startup. That turned out to be caused by the ACK / NACK messages not having proper logic for obtaining their checksum, and so that is also fixed in this PR.

A new SERIAL_UDB_EXTRA (SUE) format message F23:) is now printed after each F2 telemetry line, which contains the count of gps parse errors. The intention is that future new telemetry for SUE will be added  to the F23: message from now on, and not to the F2: message. This solves a problem with MAVLink compatibility, in that our F2: message cannot be changed in Mavlink, as it has been formerly published. Any changes to F2, would render some mavlink ground control stations as rejecting our F2 SUE over mavlink messages.

Flight Analyzer will now check for any parsing errors, and alert the user if they are found.

### DIFF
--- a/MatrixPilot/telemetry.c
+++ b/MatrixPilot/telemetry.c
@@ -677,7 +677,7 @@ void telemetry_output_8hz(void)
 					serial_output("F2:T%li:S%d%d%d:N%li:E%li:A%li:W%i:"
 					              "a%i:b%i:c%i:d%i:e%i:f%i:g%i:h%i:i%i:"
 					              "c%u:s%i:cpu%u:"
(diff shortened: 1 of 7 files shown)


phollands:
I have made a change in telemetry.c and re-pushed / force updated the branch. The printout of the magnetometer fields is now changed back to a signed integer from an unsigned integer.

PR #150, opened 2017-09-04
re-enable the build script with SILSIM and XPlane plugin targets as w…

Inspirati:
…ell as fix and cleanup the SILSIM local makefiles


phollands:
Robert,

Thanks very much for fixing the build scripts.

The buildall script runs. And it also build in the MatrixPilot-SAIL directory with "make". It creates executiables of  matrixpilot.out and silcat.out .
I would prefer that they were called  just matrixpilot and silcat.  I enclose a patch file to change the names of the executeables under *NIX, which if you find it acceptable, please can you apply it to your branch.

[patch.txt](https://github.com/MatrixPilot/MatrixPilot/files/1274334/patch.txt)

The ./matrixpilot code still does not work correctly in Fly By Wire mode, and I'll try to take a look at that soon. Under MacOS, X-Plane 10, the application of aileron in one direction creates a full throw of the aileron and rudder with only 5% aileron applied. But in the other direction the FBW works correctly. Controls work correctly in Fly By Mode. To diagnose that, I will need telemetry, and the telemetry.c file is not picking up some variables when in SILSIM. I'll take a look at that. At that time, I'll add a build to the build-all.sh script that includes SERIAL_UDB_EXTRA, which is the one that is failing to build just now.

Best wishes, Pete

Inspirati:
hey-ya Pete,

Yes indeed, i'm not sure what led me towards the old fashioned days of
going with the .out, maybe just that i'm know old enough to be inclined
towards such 'old fashioned' ways.

right now i'm in what i'm finding to be an even more longingly old lost
haunt, of trying to keep my proposed changes C30 compatible -- and yes,
it's an ongoing truly awful experience!

the past versus the present. - i'm finally beginning to appreciate the
strength of the problem C30 versus XC suppose a different root during the
build, making portability a never ending porting problem...

could i please ask of you, as our president project leader, to propose the
question to our list, shall i continue to try to maintain such build tool?
Last i understood us, Sir William was the only one (and so such
importantly) that still went there.

if we could perhaps vote to drop it, i could clean up a lot of ugly mess...

for my design objectives continue to lay with the outcome of making this
codebase be buildable on the XRacer platform - but that's another story.

also, i've again spotted this warning whilst building SILSIM with MSVC++:

1>c:\data\github\inspirati\matrixpilot\matrixpilot\flightplan-logo.c(1060):
warning C4333: '>>' : right shift by too large

perhaps this may be the logo bearing issue?

Inspirati:
1>c:\data\github\inspirati\matrixpilot\matrixpilot\flightplan-logo.c(1060):
warning C4333: '>>' : right shift by too large amount, data loss

phollands:
Robert,
As far a support for C30, we have officially dropped support for MPLAB and the C30 compiler. We are now on MPLAB-X and the XC16 compiler. I made a not about this in the release on [main page of the Wiki](https://github.com/MatrixPilot/MatrixPilot/wiki) in October 2016, and on the News page with the 5.0 release. I also made two specific announcements about dropping support for MPLAB and the C30 compiler on uavdevboard at that time. I had no come back. Bill is using MPLAB-X these days. So the C30 compiler is officially dropped from master. I know that you have been actually supporting it in builds yourself. Thank You. That has been a nice fall back facility. After a full year of everyone using MPLAB-X, and not MPLAB, we can drop support for MPLAB and C30 in master.
Best wishes, Pete

phollands:
Superseded by PR #151

PR #151, opened 2017-09-05
SILSIM with UDB_EXTRA build fix

Inspirati:
fix build error when both SILSIM and SERIAL_UDB_EXTRA defined, add tests to build script, and quiet some warnings.


phollands:
Rob, Thanks for all of this.  I will merge these into master.
I know that there are additional changes required for SERIAL_UDB_EXTRA to work in SILSIM, and I will look at those today and put in an additional PR to review with you. 
Best wishes, Pete

PR #153, opened 2017-09-05
SILSIM: re-code for SUE telemetry to work.

phollands:
This PR is primarily to allow discussion between Pete and Rob regarding the telemetry comms architecture, and getting SILSIM working.

The immediate background here is that Ric needs to test fairly complex Logo routines. One of the fastest and most reliable methods for testing Logo is with SILSIM. It is fast to compile a new change (just over 2 seconds on my machine), and X-Plane 10, can be left running during the process. So a quick change of Logo routine to testing it in flight, is just a matter of a few seconds.

So my main immediate goal is to get the SERIAL_UDB_EXTRA telemetry working in SILSIM immediately.

Beyond that, I can see that Rob has been working diligently on trying to improve and rationalise our comms architecture. 

Rob, perhaps you could review what I've done this morning (this PR), which is based on your PR that I approved into master last night; and let me know your thoughts and comments. 

Best wishes, Pete

### DIFF
--- a/MatrixPilot/telemetry.c
+++ b/MatrixPilot/telemetry.c
@@ -454,12 +454,6 @@ static void serial_output(const char* format, ...)
 		end_index = start_index + wrote;
 		udb_serial_start_sending_data();
 	}
-
-	if (sb_index == 0)
-	{
-		udb_serial_start_sending_data();
-	}
-
 	va_end(arglist);
 }
 #endif // USE_TELELOG
--- a/Tools/MatrixPilot-SIL/SIL-serial.c
+++ b/Tools/MatrixPilot-SIL/SIL-serial.c
@@ -95,12 +95,28 @@ void sil_telemetry_input(uint8_t* buffer, int32_t bytesRead)
 // Call this function to initiate sending data to the serial port
 void udb_serial_start_sending_data(void)
 {
+	uint8_t buffer[BUFLEN];
 	int16_t c;
 	int16_t pos = 0;
+	int16_t bytesWritten;
+	
+	if (!telemetrySocket) return;
 
 	while (pos < BUFLEN && (c = udb_serial_callback_get_byte_to_send()) != -1) {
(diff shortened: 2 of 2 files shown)


Inspirati:
Looks good to me, that's pretty much what I was thinking for what might fix it. Question is though, does it now work?

One issue is with this arrangement of code. Besides my being picky with C code style consistency, this actually won't build with certain versions of VC++ (like the one i was using last night, since our last MAVLink update brought this issue in with it again). May i recommend moving the 'if (...) ..' to after the variable declarations.

        if (!telemetrySocket) return;
+	uint8_t buffer[BUFLEN];
 	int16_t c;
 	int16_t pos = 0;
+	int16_t bytesWritten;

I'll aim to get a PR up addresses the few instances in the MAVLink code which do this too.

phollands:
I will change it as suggested. And yes it is working for Silcat to pick up
the SUE telemetry.



On 5 Sep 2017 11:18, "Robert Dickenson" <notifications@github.com> wrote:

Looks good to me, that's pretty much what I was thinking for what might fix
it. Question is though, does it now work?

One issue is with this arrangement of code. Besides my being picky with C
code style consistency, this actually won't build with certain versions of
VC++ (like the one i was using last night, since our last MAVLink update
brought this issue in with it again). May i recommend moving the 'if (...)
..' to after the variable declarations.

    if (!telemetrySocket) return;


   - uint8_t buffer[BUFLEN];
   int16_t c;
   int16_t pos = 0;
   - int16_t bytesWritten;

I'll aim to get a PR up addresses the few instances in the MAVLink code
which do this too.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/MatrixPilot/MatrixPilot/pull/153#issuecomment-327134252>,
or mute the thread

phollands:
I've made Rob's suggested change and pushed it up to the branch.

PR #155, opened 2017-09-05
UDB_Logo_Editor: Fix calculation of angles when using atan2().

phollands:
I am going to ask Ric, to test this change. The returned angles were previously negative for headings between 270 degrees and due North. They are now positive values between zero and 360 degrees.

I have tested the following snippet which was previously faulty but now works correctly in the edtor:-

PEN_UP
    USE_CURRENT_POSITION
    IF_LT(ANGLE_TO_HOME, 180)
        WEST(200)
    ELSE
        EAST(200)
    END

The following also works correctly:-

IF_GT(REL_ANGLE_TO_HOME, 0)
	WEST(300)
ELSE
	EAST(300)
END

As shown in these two screenshots where the angle of the plane relative to the origin varies just slightly from just 0 degrees to just above 0 degrees.

screen shot 2017-09-05 at 14 38 43

screen shot 2017-09-05 at 14 38 55

### DIFF
--- a/Tools/UDB_Logo_Editor/udb_logo_editor.html
+++ b/Tools/UDB_Logo_Editor/udb_logo_editor.html
@@ -172,8 +172,10 @@
 		var relPlaneAnglePt = absoluteToRelative(planeAnglePoint);
 		var dx = relPlaneAnglePt.we-relPlanePt.we;
 		var dy = relPlaneAnglePt.sn-relPlanePt.sn;
-		var angle = Math.atan2(dy, dx);
-		planeAngle = 90 - (angle * 180/Math.PI);
+		var angle = 90 - (Math.atan2(dy, dx) * 180 / Math.PI);
+		if (angle < 0)
+			angle = angle + 360;
+		planeAngle = angle;
     }
 
     function savePlanePoint(){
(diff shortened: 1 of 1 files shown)


phollands:
I an email to me, Ric stated:-
"Pete,
Thank you very much.
With the flight program you mentioned it works now.
I did a test with the following program too, but it seems it is not interpreting REL_ANGLE_TO_HOME correctly."

I agree with Ric. And then, looking back through uavdevboard, I see that [Kees had a discussion in June 2014](https://groups.google.com/d/msg/uavdevboard/zVg2JYZ1MUY/tGa_vjTdnCMJ), to say that the REL_ANGLE_TO_HOME, and REL_ANGLE_TO_GOAL had the wrong sense for positive and negative. So he created [a patch, which Robert put into master.](https://github.com/MatrixPilot/MatrixPilot/commit/9883c34ba895d3b12a58f66b8f16e807232b17cd) But at the time the Logo editor was not also updated.

So I have now also updated the Logo Editor with a second commit above, so that the Logo simulation follows the same convention as the firmware in the plane:-  REL_ANGLE_TO_HOME is positive if the angle is to the right of the plane. And likewise for REL_ANGLE_TO_GOAL.

phollands:
This PR is awaiting final testing from Ric.

PR #156, opened 2017-09-06
GNSS (Global Navigation Sattelite System) metrics

phollands:
This PR is primarily for people that fly in places where the GNSS accuracy may be obscured by mountains and buildings. It has two features:-

- The user can specify that the "GPS Acquired" state does not occur until the following metrics are met:-
      o  HDOP exceeds a value set in options.h
      o SVS exceed as value set in options.h
      o And for the Ublox only, VDOP exceeds a value set in options.h
   The usual standby by timer is also in effect, so the GNSS device always has to settle down for a period before "GPS iacquired."
- The VDOP values are now embedded in the SERIAL_UDB_EXTRA telemetry. This will help us to measure whether VDOP is significantly effected when the plane is low down in Alpine valleys, with mountains high all around.

I have started to test the scheme, and I invite comments from anyone.

### DIFF
--- a/Config/Cessna/options.h
+++ b/Config/Cessna/options.h
@@ -74,6 +74,15 @@
 #define GPS_TYPE                            GPS_STD
 //#define DEFAULT_GPS_BAUD                    57600   // added for GPS_NMEA support
 
+////////////////////////////////////////////////////////////////////////////////
+// You can specify a level of good GNSS reception before MatrixPilot accepts "GPS ACQUIRED".
+// You can generally leaves these lines at their default values. A value of zero switches off the check.
+// The VDOP parameter is only available for Ublox GNSS devices. It is ignored for other GNSS units.
+// The metrics are not used by HILSIM or SILSIM.
+
+#define GNSS_HDOP_REQUIRED_FOR_STARTUP       20  //  Horizontal Dilution of Precision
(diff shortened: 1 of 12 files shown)


phollands:
I will add some code so that HILSIM / SILSIM does not run the metric checks, as it really is not required in that situation.

phollands:
In a flight test today, I had slightly different results to what I expected. The normal waggles starts when the standby timer for the gps has decremented to zero. However, the waggles then continue until the all the GNSS metric requirements have been met. So there was a lot of aileron waggling. This is was nice for me in testing, as I immediately knew that the GPS was half acquired but not completely acquired for the metric that I had set-
```
#define GNSS_HDOP_REQUIRED_FOR_STARTUP       10  //  Horizontal Dilution of Precision
#define GNSS_VDOP_REQUIRED_FOR_STARTUP	     60  //  Vertical Dilution of Precision
#define GNSS_SVS_REQUIRED_FOR_STARTUP	      8  //  Number of Sattelites in View
```
So I quite like the feature from the point of view of starting up, but it would not be good for an in flight re-boot from a power glitch, as the plane would be quite uncontrollable. 

BTW: It was a great flight with my flying wing. Master is flying well.

phollands:
Commit 679d492 Flight tested OK. Normal initialization of Ublox unit after GNSS metrics satisfied, with normal "GPS Acquired" waggles of the ailerons.

PR #157, opened 2017-09-06
Fix to allow for bigger interrupt programs.

KeesGuijt:
Before this the number of instructions in interrupts could not exceed MAX_INSTRUCTIONS_PER_CYCLE.

Simply making MAX_INSTRUCTIONS_PER_CYCLE larger would cause 40(Hz) * x instructions per seconds

This fix: the first cycle will run the first MAX_INSTRUCTIONS_PER_CYCLE instructions,
but will continue with the next cycle to run the instructions that follow the first batch. If no CLEAR is encountered, the routine will run to its end and start over.
Interrupt routines are only run while the main program is waiting for a fly command to finish.

### DIFF
--- a/MatrixPilot/flightplan-logo.c
+++ b/MatrixPilot/flightplan-logo.c
@@ -466,7 +466,7 @@ void flightplan_logo_update(void)
 		}
 		else
 		{
-			if (interruptStackBase)   //if not in-progress
+			if (interruptStackBase)   //if in-progress
 			{
 				//cleanup uncompleted interrupt
 				instructionIndex = logoStack[interruptStackBase].returnInstructionIndex+1;  //support both end by main and end by END SUBROUTINE
--- a/MatrixPilot/flightplan-logo.c
+++ b/MatrixPilot/flightplan-logo.c
@@ -443,19 +443,36 @@ void flightplan_logo_update(void)
 	}
 
 	// otherwise run the interrupt handler, if configured, and not in-progress
-	if (interruptIndex && !interruptStackBase)
+	if (interruptIndex)
 	{
-		if (logoStackIndex < LOGO_STACK_DEPTH-1)
+		//if not arrived
+		if (tofinish_line >= WAYPOINT_PROXIMITY_RADIUS) // not crossed the finish line
 		{
-			logoStackIndex++;
-			logoStack[logoStackIndex].frameType = LOGO_FRAME_TYPE_SUBROUTINE;
-			logoStack[logoStackIndex].arg = 0;
-			logoStack[logoStackIndex].returnInstructionIndex = instructionIndex-1;
-			instructionIndex = interruptIndex+1;
-			interruptStackBase = logoStackIndex;
+			if (!interruptStackBase)   //if not in-progress
+			{
+				if (logoStackIndex < LOGO_STACK_DEPTH-1)
+				{
+					logoStackIndex++;
+					logoStack[logoStackIndex].frameType = LOGO_FRAME_TYPE_SUBROUTINE;
+					logoStack[logoStackIndex].arg = 0;
+					logoStack[logoStackIndex].returnInstructionIndex = instructionIndex-1;
+					instructionIndex = interruptIndex+1;
+					interruptStackBase = logoStackIndex;
+				}
+			}
 			process_instructions();
 			navigate_set_goal_height(turtleLocations[PLANE].z);
 			lastGoal.z = turtleLocations[PLANE].z;
+		}
+		else
+		{
+			if (interruptStackBase)   //if not in-progress
+			{
+				//cleanup uncompleted interrupt
+				instructionIndex = logoStack[interruptStackBase].returnInstructionIndex+1;  //support both end by main and end by END SUBROUTINE
+				logoStackIndex =  interruptStackBase - 1;
+				interruptStackBase = 0;
+			}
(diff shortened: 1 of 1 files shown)


phollands:
Kees, 

Please can you write about how you've been testing this. Have you been flying with this modification already ?

Please can you link to an example flight plan in Logo that has needed this fix, so I can understand the context a little better.

In the longer term, do we need a different architecture for monitoring flight events in Logo, or do you think that this is sufficient ? e.g. Are we doing GeoFencing the best way by using "Interrupts", or really do we need a better architecture ?

I  am going to invite  Ben to comment / review, as he is the designer and author of our Logo implementation.

Thanks for your contributions, and I hope you've been having some fun Autonomous Thermaling flights this season.

Best wishes, Pete

KeesGuijt:
Pete,

i've been trying to make more use of interrupts in my LOGO script. Many attempts to make changes failed, when i tested the code in Hilsim.
My LOGO code consists of fly commands that last about 1 second. In between the fly commands a number of checks are performed.
https://github.com/KeesGuijt/MatrixPilot/blob/a562100c69452e61298774945da0053ea1852fb5/Config/flightplan-logo.h#L1143

For example:
			IF_LT(ALT, 10)  //below: auto takeoff / hand launch with motor on in Autonomous mode
			    DO (TAKEOFF)             //keep level when low
			END
			FD(10)

Problem is that at some points i need upto 12 checks, and preferably like this:

	TO (INT_CRUISE)  //interrupt routine
		DO (CHECK_PILOT)
		DO (CHECK_GF)
		DO (CHECK_LAND)
		DO (CHECK_SINK)
		DO (CHECK_THERMAL)
		DO (CHECK_HIGH)
		DO (CHECK_LATE)
		DO (CHECK_OVERSHOOT)
		DO (CHECK_MOTOR)
		DO (CHECK_SOFT_GF)
		DO (CHECK_TAKEOFF)
		DO (CHECK_TEST_MODE)
	END
	END

When i would add the checks to the main loop, the code would often take too long, causing the plane to fly past the turtle. The plane then does an unwanted 360.
Adding the checks to interrupt routines did'nt help either (without the fix).
Of course, LOGO was never intended for this kind of abuse, i realize that.

Testing was done with about 20 hilsim flights, and one live flight. 

I'm happy with this architecture for now, my LOGO script is cleaner now, and easier to read too.
Using LOGO is not allways easy, but it is a nice way to describe the high level logic used. For the tough lower level stuff, i wrote a bunch of custom commands in c.
[https://github.com/KeesGuijt/MatrixPilot/blob/2339b2e4bd3a7ef1ff0f8b55c263d4efe5da5062/MatrixPilot/flightplan-logo.c#L82](url)

Have a closer look at the geofencing code here  [https://github.com/KeesGuijt/MatrixPilot/blob/2339b2e4bd3a7ef1ff0f8b55c263d4efe5da5062/MatrixPilot/flightplan-logo.c#L1700](url) I wrote it with the help of my collegue Philip. 

I haven't had many chances to fly this summer, but hopefully i can make a video to show the code in action.

All the best,
Kees

MatrixPilot/flightplan-logo.c
```diff
@@ -443,20 +443,37 @@ void flightplan_logo_update(void)
...
+					logoStackIndex++;
+					logoStack[logoStackIndex].frameType = LOGO_FRAME_TYPE_SUBROUTINE;
+					logoStack[logoStackIndex].arg = 0;
+					logoStack[logoStackIndex].returnInstructionIndex = instructionIndex-1;
+					instructionIndex = interruptIndex+1;
+					interruptStackBase = logoStackIndex;
+				}
+			}
 			process_instructions();
 			navigate_set_goal_height(turtleLocations[PLANE].z);
```
tripoverzero: Should this line and the following line be here?  I think they might make more sense running after the interrupt completes.

MatrixPilot/flightplan-logo.c
```diff
@@ -443,20 +443,37 @@ void flightplan_logo_update(void)
...
 			lastGoal.z = turtleLocations[PLANE].z;
 		}
+		else
+		{
+			if (interruptStackBase)   //if not in-progress
```
benjie-git: I think this comment should read:
// if in-progress

benjie-git:
Am I right that this change could allow a bad interrupt with an accidental infinite loop, to block the actual logo program from ever getting run?  That is probably ok, as long as we make that clear.

MatrixPilot/flightplan-logo.c
KeesGuijt: Ben,

These lines where already there. They are run after the interrupt completes, but now also each time after MAX_INSTRUCTIONS_PER_CYCLE was reached with larger interrupt programs
But i'm not completely shure about the purpose of those two lines, because this function is also performed in process_instructions()?

Regards,
Kees.

KeesGuijt:
Ben,

 "Am I right that this change could allow a bad interrupt with an accidental infinite loop, to block the actual logo program from ever getting run? That is probably ok, as long as we make that clear. "
	 
I suppose an infinite interrupt loop should be no problem. Every cycle MAX_INSTRUCTIONS_PER_CYCLE interrupt instructions will be run, then MAX_INSTRUCTIONS_PER_CYCLE actual logo program instructions.   (see line 488 or 495).

Kees

PR #159, opened 2017-09-11
SILSIM: Fix for FBW mode to operate correctly.

phollands:
Rob, 

I changed the builtins.h file (which only is used for non dspic devices) this morning to better reflect my understanding of the size of the integers required for each of the parameters, and also the size of the returned values.  Fly By Wire Mode is now running correctly in SILSIM on my setup, with X-Plane 10, in that, on the ground, the Ailerons / Rudder no longer suddenly flip to nearly full throw on the application of "5% aileron" with the keyboard.  The J and Keys   correctly move the ailerons and rudder in fly by wire mode with this patch applied.

Please can you test this change, on your PC  and let me know if the fix works for you.

I know that casting can be tricky, and that this change needs some careful reviewing. It may well effect any other ports to non dspic processors. (e.g. ARM processors.).

Best wishes, Pete

### DIFF
--- a/libUDB/builtins.h
+++ b/libUDB/builtins.h
@@ -12,11 +12,11 @@
 // Fake dspic builtin ASM calls for Non dspic devices (e.g. Arm devices))
 // Microchip dspic controllers use __builtin__ calls to engage special fast instructions in the hardware
 // This file is not included in builds for Microchip dspic devices
-#define __builtin_mulss(x,y) ((( int32_t)(x))*( int32_t)(y))
-#define __builtin_mulus(x,y) (((uint32_t)(x))*( int32_t)(y))
-#define __builtin_mulsu(x,y) ((( int32_t)(x))*(uint32_t)(y))
-#define __builtin_muluu(x,y) (((uint32_t)(x))*(uint32_t)(y))
+#define __builtin_mulss(x,y) (( int32_t)((( int16_t)(x))*( int16_t)(y)))
+#define __builtin_mulus(x,y) (( int32_t)(((uint16_t)(x))*( int16_t)(y)))
+#define __builtin_mulsu(x,y) (( int32_t)((( int16_t)(x))*(uint16_t)(y)))
+#define __builtin_muluu(x,y) ((uint32_t)(((uint16_t)(x))*(uint16_t)(y)))
 #define __builtin_divud(x,y) ((uint16_t)(((uint32_t)(x))/(uint16_t)(y)))
-#define __builtin_divsd(x,y) (( int16_t)((( int32_t)(x))/(int16_t)(y)))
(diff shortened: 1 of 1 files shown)


Inspirati:
G'day Peter,

Wow, I noticed that during my weekends debugging effort and initially
thought, 'gotcha, that's gotta be it!'. However when I changed it out and
performed another test run, the rudder problem was still apparent. I then
moved on to checking if the bug existed as far back as release 4.1 and
confirmed it does not. I then testing the helicalTurns branch and
discovered it is featured there. From then I spent many hours trying to
work out how I could further isolate the introduction point, but got lost
in git semantics and discovering that our googlecode repository doesn't
seem to be available anywhere. I thought google was just going to make it
read only, and keep it around for posterity..

Anyhow, I now suspect I must have stuffed up my test and so shall perform
it again to confirm you findings. Well done, but it bothers me that I may
have found it, and then made a mistake in confirming such )-:

Thanks, Robert.

Inspirati:
Confirmed. Well done, a nasty bug well squashed.

The 'fix' I attempted to implement, which was meant to be swapping in the inline functional versions as such:

static inline int32_t  __builtin_mulss(const int16_t p0,   const int16_t p1)   { return (p0 * p1); }
static inline int32_t  __builtin_mulus(const uint16_t p0,  const int16_t p1)   { return (p0 * p1); }
static inline int32_t  __builtin_mulsu(const int16_t p0,   const uint16_t p1)  { return (p0 * p1); }
static inline uint32_t __builtin_muluu(const uint16_t p0,  const uint16_t p1)  { return (p0 * p1); }
static inline uint16_t __builtin_divud(const uint32_t num, const uint16_t den) { return (num / den); }
static inline int16_t  __builtin_divsd(const int32_t num,  const int16_t den)  { return (num / den); }

also works now that i've tried it again. So I must have missed something when doing the build. Probably got my branches mixed up and changed the wrong one. Doh!!

phollands:
Rob, If the use of "static inline" is viewed as a better and more efficient implementation, then I'm happy to go with that. 
Is that what you would prefer ? If so, I'll change my commit to use your code sample, and re-test it.
Best wishes, Pete

phollands:
Improved Syntax, tested OK in X-Plane 10, with ./matrixpilot on Imac.

MatrixPilot:
Yeah I'm not sure which way may be 'better', so not fussed for now. Can't
recall why I implemented the function versions, and more importantly have
no idea how I managed to get the broken macro versions merged in. I believe
a common school of thought is that macros are bad, and I also suspect the
likes of Scott Meyers would say go the the inline functions.

Oh! it just came to me. The function versions will give the compiler a
chance to throw an error/warning when called with incompatible types,
where-as the macros will silently cast away. So it's all about type safety.

PR #161, opened 2017-09-13
HILSIM XPlane 64-bit build

Inspirati:
Fix SILSIM magnetometer handling, improve makefiles, and add a 64-bit build of the XPlane plugin for linux.


phollands:
Rob,

The HILSIM 32 and 64 bit execute-ables for X-Plane build nicely on Ubuntu 16.04. I am not able to test the plugins them as my Ubuntu is running under Parallels.
The SILSIM version of MatrixPilot build (with two warnings previously discussed), and run just fine. I am able to conect the Ubuntu MatrixPilot via Parallels to X-Plane 10 running natively on the Mac. Surprisingly the frame rate for X-Plane 10 is up at 50 to 60 / second which is good. The simulation is working fine, given that Ubuntu and Macos are running together on the same hardware.

I have yet to test how this PR works for Macos and WIndows 10. 

screen shot 2017-09-13 at 11 24 10

Inspirati:
I'm testing on Win7 now and i mucked up with my device-HIL.mk, fixing now..

Inspirati:
interesting from your screenshot i see a bunch of warnings that don't
appear in my builds, such as:

warning: "ID_LEAD_PILOT" redefined

are you working with a clean checkout, or something more customised?

phollands:
You can ignore those warnings in the screenshot. I had merged my options.h file for the Easy Star RC model  into your branch HILSIM_XPlane64 options.h and I made a mistake by leaving 3 lines re-defines about the lead pilot etc. They were not important to the test.

Inspirati:
okay cool.

btw, have you tried running the build-all.sh script on your Win10 VM?

i'm curious to know what it does on a reasonably clean installation.

phollands:
Can you send me a link to the location for instruction on how to create the same build environment as yourself under Windows ?  I have tried [using MSYS2 ](http://www.msys2.org/)(Mingw64 MINGw32) as that looked like a vibrant development community, but it does not work well for our Makefiles.

phollands:
Rob, I did get Windows 10 64bit MatrixPilot working of sort. It will fly X-Plane 10; but the Reset function exits the program, rather than re-starting the simulated UDB.  I used Visual Studio 2017 and upgraded the project (a button click). I had to undefine two places for snprintf in MatrixPilot-SIL and HILSIM_XPlane. I enclose a screenshot of the diffs below. All the other diffs are my preferred optoins.h, and SIL-Config.h points that the IP addess of my Mac where the X-Plane simulator is actually running (i.e. only MatrixPilot-Sil.exe is running on Windows 10).

It would be nice to agree a definitive build environment that is fairly easy for users to get working.

Best wishes, Pete

screen shot 2017-09-13 at 17 08 02

screen shot 2017-09-13 at 17 18 31


And X-Plane 10 on the Mac, being flown from Windows 10 in fly by mode, in a right hand turn. (Sorry I missed the very bottom screen).

screen shot 2017-09-13 at 17 03 21

phollands:
I enclose a screenshot of the build-all.sh process using mingw64. It just hangs, in that there is no more activity after the make begins.

screen shot 2017-09-13 at 22 29 32

Inspirati:
Hi Peter,

How long did you leave it 'hung'? The builds are a *lot* slower on Windows
than you would be used to under Darwin. At the point displayed, the build
should be occurring, and in lieu of any warnings being thrown, you won't
see anything until completion. The XPlane plugin is not a very large build,
but did I mention *slow*. During my test runs last night, the entire script
was taking well over half an hour! Mind you, I've never installed MinGW64,
so we may well have a problem, and it may be even slower. If you set an
environment variable SET V=1 before running the script, i'll be more
verbose.

As of my PR from yesterday, the XPlane makefile (the one in the
HILSIM_XPlane directory) aims to build both the 32 and 64 bit targets. I
have already wondered how that may play out on other platforms. We both
already discovered that some additional headers need to be installed in
order to have 64 bit unix be able to build the 32 bit target. I noted the
same issue on Windows last night, but am yet to address it. I'm not sure
that the makefile which runs with the build-all script (the one in the root
directory) particularly specifies a 32 or 64 bit target, and as such expect
that it would build the default - 32 on 32 bit system, 64 on 64 ..

Unfortunately there is no methodology for recreating my Windows build
environment, as it's evolved over like the past eight years. There's more
junk and mess on that machine than I care to admit too. I would've hoped
MSYS would do the trick, however I have added numerous other unixie
utilities along the way, generally in support of my other work with other
projects.

I suspect the SILSIM exit on reset will have to do with the joy of the
_exec(..) function. I've had my share of grief with it's various compiler
versions incarnations before. Did SILSIM display the string "Failed to
reset UDB ..." on exit, or just crash out?

And as for the snprintf defines, if we want to support older versions of
VC++, we may need to add a conditional compilation path.

Thanks for your testing and report.

Best wishes, Robert.

phollands:
Rob,

Your PR works well for Ubuntu 16.04 in that it builds matrixpilot, silcat, and the HILSIM plugin. I cannot myself test the HILSIM plugin in X-Plane 10, as I don't currently have a working native Ubuntu computer.

Please can you confirm the version of Ubuntu that you used for testing both matrixpilot and the HILSIM plugin attached to X-Plane 9 ?

Before merging this, I would like to note the longer term vision here, and where I think we are at today, with testing.

SILSIM can make it easy for a new user to test out  MatrixPilot with X-Plane 9 and 10, without purchasing a board.  SILSIM is also good for testing out new Logo flight plans, because the turn around time for compiling MatrixPilot natively on a 32, or 64 bit platform is typically about one to two seconds.

Summary of current testing:-
MatrixPilot-SIL and Silcat:- 
   Ubuntu 16.04. Will build with a simple "make" 
   MacOS Sierra: Builds with a simple "make" Runs OK.
   Windows: builds with caveats:- I have settled on using [MSYS2](http://www.msys2.org/) as the gcc build environment. It seems to be  the best supported gcc environment under Windows. It also potentially allows cross compiles from other OS environments. e.g. Ubuntu. At the moment, I do an initial install as outlined [here](http://www.msys2.org/), and then add the development environment with
 pacman -S --needed base-devel mingw-w64-i686-toolchain mingw-w64-x86_64-toolchain git
The makefile has to edited to make "clean" work from mingw64 or 32 shells, as they use "rm" and not "del" for their commands. 
The resulting binary has to be run from within a normal dos cmd window, and not from within Mingw64 or MingW32. If it is run inside the MingW64 / 32 shells, then matrixpilot will connect to X-Plane, but the terminal I/O does not work correctly in that keyboard input does not work. I expect the fix is simply to select the correct CRT library for mingw32 / 64 windows. But anyway, the binary works from the cmd window.So there is work to be done with the Windows environment but I can see a way through.

HILSIM Plugin: Ubuntu: Rob's testing.
MacOS: I can compile the plugin using "make -f Makefile.macos" after editing the location of the plugins directory in the Makefile. The main makefile does not work.
Windows: I have not been able to make that work.

Best wishes, Pete

PR #162, opened 2017-09-14
LOGO WIND_FROM_ANGLE

KeesGuijt:
Add a LOGO "system value" to read the wind direction, as discussed [here](https://groups.google.com/forum/#!topic/uavdevboard/CUopQNA2Ouo)

### DIFF
--- a/MatrixPilot/flightplan-logo.c
+++ b/MatrixPilot/flightplan-logo.c
@@ -641,8 +641,8 @@ static int16_t logo_value_for_identifier(uint8_t ident)
 		case WIND_FROM_ANGLE: // wind from in degrees 0-359, 0 = North
 		{
 			int16_t angle = get_angle_to_point(estimatedWind[0], estimatedWind[1]);
-			while (angle < 0) angle += 360;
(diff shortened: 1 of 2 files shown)


MatrixPilot/flightplan-logo.c
```diff
@@ -637,6 +638,14 @@ static int16_t logo_value_for_identifier(uint8_t ident)
 		case WIND_SPEED_Z: // in cm/s
 			return estimatedWind[2];
 
+		case WIND_FROM_ANGLE: // wind from in degrees 0-359, 0 = North
+		{
+			int16_t angle = get_angle_to_point(estimatedWind[0], estimatedWind[1]);
+			while (angle < 0) angle += 360;
```
benjie-git: I think these while loops should just be if blocks, unless there's a way get_angle_to_point() can return under -360 or over 720.

KeesGuijt: Ben,

Ok, i see that most of the similar code uses if's, i'll change that. Thanks. 
  
Best regards,
Kees

PR #163, opened 2017-09-17
fixed a bug that triggered inverted flight during 4 G turns. the caus…

wjpremerlani:
…e was an unsigned value that was passed as the second argument to __builtin_divsd

### DIFF
--- a/MatrixPilot/helicalTurnCntrl.c
+++ b/MatrixPilot/helicalTurnCntrl.c
@@ -313,14 +313,14 @@ void helicalTurnCntrl(void)
 	if (desiredTilt.WW > (int32_t)2 * (int32_t)RMAX - 1)
 	{
 		desiredTilt.WW = (int32_t)2 * (int32_t)RMAX - 1;
-		accum.WW = __builtin_mulsu(desiredTilt._.W0, GRAVITYCMSECSEC);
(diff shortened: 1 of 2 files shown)


MatrixPilot/helicalTurnCntrl.c
```diff
@@ -313,14 +313,14 @@ void helicalTurnCntrl(void)
 	if (desiredTilt.WW > (int32_t)2 * (int32_t)RMAX - 1)
 	{
 		desiredTilt.WW = (int32_t)2 * (int32_t)RMAX - 1;
-		accum.WW = __builtin_mulsu(desiredTilt._.W0, GRAVITYCMSECSEC);
+		accum.WW = - __builtin_mulsu(desiredTilt._.W0, GRAVITYCMSECSEC);
```
phollands: Bill, Please can you explain the reason for adding the negation at line 316. Pete

wjpremerlani: Peter,
There we actually two errors. The sign of tilt is opposite to the turn rate. You can see that in the code for the unsaturated case. At first I thought that was the source of the bug, but it turned out during my first few test flights the saturation never was triggered. - bill

PR #165, opened 2017-09-19
Fixed a bug in normalization routines that caused 4 G lateral acceler…

wjpremerlani:
…ation turns to invert. Also fixed a sign error in the turn rate computation at 4 G.


phollands:
Passes Testing in HILSIM with Ric's Logo Flight Plan.

PR #152, opened 2017-09-05
Move SILcat to own directory and add to scripted build targets. Move …

Inspirati:
…UDBSocket* to common location for all three targets which include them, adjusting the build system to suit.

PR #174, opened 2017-10-15
X-Plane plugin: Configure makefile and code for Windows (MSYS2)

phollands:
I now have the plugin compiling without warning or errors on WIndows 10  (MSYS2), MacOS (High Sierra, using X-Code Gcc tools), and Ubuntu 16 .04.

Rob, I intend to do some more testing myself, but would be grateful if you have time, for you to look over some these changes and comment where useful. 

I intend to write up a detailed example of installing the full SILSIM suite for new users.
i.e. X-Plane 10 installation and configuration, installing a development environment; installing the HILSIM_Setup.txt file and HILSIM plugin for X-Plane; Compile and run MatrixPilot (run the "fly script").


Inspirati:
Nice work Pete,

I notice a few of the changes are familiar to me from my PR the other month which I never got back to. Did you aim to include everything from that PR or should I revisit it for an audit?

I would remove the .cdep and .cppdep files are they are compiler generated an as such do not belong in the repository. Indeed the build directory should not exist at all.

Curious that lines 335 and 336 have snuck back in to UDBSocketWin.c

OSOBJS could be dropped from the makefile as it is no longer used. Not quite sure why I went with that, but if an issue arises it'll be simple enough to reincorporate it.

Having read [this](https://www.quora.com/What-is-the-difference-between-the-GCC-compiler-flags-fPIC-and-fpic-When-should-one-use-one-the-other-or-both) last week, I wonder if we should also consider testing a change to our build flags. I suspect if this ever did raise its ugly head, it would result in an awfully difficult bug to track down. Probably not an issue for the plugin, however our other targets may be susceptible so figured I'd mention it while I remember.

phollands:
I have left PR #152 "Move SILCAT to own directory" for you to adjust for the git conflicts that are there. I did not use your PR Makefiles for my PR which is focusing on getting a baseline of working code on all three OS platforms.

The whole HILSIM build directory and it's tree (including .cdep etc) is now ignored by .gitignore setting of  build/** .

If the functional testing of the plugin works on Window 10, MacOS High Sierra, and WIndows, then I am minded to merge this PR. That will then give you a clean foundation on which make the changes for moving silcat to a new directory, if you still want to make that change.

I'll take a look at the lines in UDBSocketWin.c .

phollands:
I see what you mean now about the .cdep and .cppdep files being committed. That was a mistake. I've taken them out.

phollands:
Although the plugin builds without error on all three OS platforms, the Windows plugin is not loading into X-Plane. Failing for now in Windows. Ubuntu and MacOS plugins are working.

wjpremerlani:
Peter, if this is any help, I note that Xplane 10 on my machine has two executables, X-Plane-32bit.exe and X-Plane.exe. Both of them will work ok stand alone, but on my Windows box, a 64 bit machine running windows 7, only X-Plane-32bit will play nice with the HILSIM plugin. For Xplane 9, there is only one executable, X-Plane.exe, and it runs just fine with HILSIM.
I also note that there earlier versions of the plugin hardly ever worked for me with either Xplane 9 or 10, it had trouble establishing the communications link. Eventually Ben made some changes that fixed that. Now HILSIM works for me the first time, every time I run either Xplane9 or 10.

phollands:
Both 64bit and 32Bit X-Plane under Windows 10 is now working. The solution was to start X-Plane from within the Mingw32 or Mingw64 command line terminal windows for X-Plane 32 bit or X-Plane 64 bit respectively. So there is some dynamic library that is required at run time, that I have not yet tracked down, which is provided as part of the Mingw environment. I do want X-Plane to be able to be started from a double click in the file manager (which you cannot do with the branch as it is now). So further work is required. But I'm making progress slowly. 

Rob, I have taken out the lines for OSOBJS from the makefile because, as you pointed out, they were not needed. Likewise the linking to -lmesa was not needed (OpenGL). Iff you see anything else that should be changed, please let me know.

phollands:
Just a note, to say that while travelling, I tried this makes file out on a older download of X-Plane 10 on MacOS El Captian. It failed. It only works correctly when the X-Plane demo was updated to the latest version (Version 10.51). Clearly some of the plugin functionality was broken in the older versions of X-Plane 10, and in particular some of the support for 64 bit computing was not in the DLL / Frameworks in the Resources/plugin directory. (seen by inspection of that directory).

phollands:
I used Process Explorer from Microsoft to see which DLL's where being dynamically loaded by the X-Plane HILSIM plugin from the mingw64/bin plugin.  It was using the winpthread dll. So I have statically loaded that DLL, so that it will always be found whether or not mingw64/bin is in the PATH.

PR #175, opened 2017-10-16
DCM improvements

wjpremerlani:
Improved the core algorithm for computing attitude by using a gps-less algorithm to compensate for centrifugal effect and by using the tilt row of the direction cosine matrix as the reference for renormalization.

### DIFF
--- a/libDCM/rmat.c
+++ b/libDCM/rmat.c
@@ -144,9 +144,6 @@ static fractional errorRP[] = { 0, 0, 0 };
 static fractional errorYawground[] = { 0, 0, 0 };
 static fractional errorYawplane[]  = { 0, 0, 0 };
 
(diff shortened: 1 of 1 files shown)


phollands:
Bill has done Bench Testing of the board to confirm the accuracy of the IMU with these changes. He has also done HILSIM. I have done SILSIM tests and a HILSIM test where I bring back the X-Plane 10 pitch and roll from the simulator, and compare that to the DCM. I enclose an example below.
Bill and I have discussed the results and agreed to merge this commit.

imu_test_branch_dcm_improvments_graph

The oscillations in the DCM (right hand side of graph) occur at highly aggressive bank angles of 80 degrees when the elevator is increased (creating a very fast rotation, and increased angle of attack). This new algorithm actually has a better performance in these conditions than the old / previous code base.

PR #177, opened 2018-05-05
Flight control telemetry

phollands:
I am ambivalent about adding these changes to master, as I have not seen much interest in them so far. But I'm going to post this Pull Request here for now, to see if anyone actually would like this facility.

This set of changes allows the effects of PID flight control gains to be analyzed in more detail; either using a csv file / spreadsheet, or using a [graphing tool within an inflight video. ](https://vimeo.com/264754315). I have used "racerenderer" and I could add the templates for racerender to pick up the data from the csv file, and to layout the graphs on the video. I think it is quite useful for seeing whether the Tilt error, Rate of rotation error, or Desired Rotation, or dynamic Trim, is responsible at any moment, for moving the ailerons, rudder or elevator.

I have tested this on my sonic mini flying wing. And now hope to test it also on my Alpha Axion trainer (standard airframe).

The only downside of this request, is that the F23: format statement in the SUE telemetry will be longer, and so your SD card in the OpenLog will need to be fast enough to keep up with saving the data. Mine does. But I know that historically, some people have slower SD cards, and might start to drop data.

### DIFF
--- a/Config/options.h
+++ b/Config/options.h
@@ -794,6 +794,19 @@
 #define ID_LEAD_PILOT                       "Not Defined"
 #define ID_DIY_DRONES_URL                   "http://www.diydrones.com"
 
+///////////////////////////////////////////////////////////////////////////////////
+// Aid to Flight Analysis of Delta Wing Aircraft. This is an optional setting which
+// allows for minimizing drag of the aircraft from commanded elevator drag (in the elevons).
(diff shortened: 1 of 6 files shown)


elgarbe:
Peter, I'm very interesting on this work!!! I'm not flying too much this days, but I will use your code/template for next flight.
Thank you very much ir!

Inspirati:
mmm, due to your earlier analysis, i would not have voted for this either!,
but i figure no harm done..

PR #178, opened 2019-06-10
flan.pyw: Find C pre-processor with a more general search 

phollands:


phollands:
This change to flan.pyw is made to ensure that flan can find the C pre-processor (required for processing the waypoints.h files) for future version numbers of the XC16 compiler when used under Windows.

PR #181, opened 2019-08-08
Sonar / Lidar Driver

phollands:
These changes add to the work started by Robert, to take code from an SVN branch called MatrixPilot_Sonar (branch was circa 2012), and move it into master. I have flight tested these commits on the UDB5 today. There is some additional features (e.g. allowing LOGO to see sonar height above the ground) that still could be ported over.  These commits provide distance to target, and distance above the  ground in the telemetry. Flan.pyw already knows how to include this information in the standard csv spreadsheet.


phollands:
I've added ALT_AGL to log as Altitude Above Ground Level to create a variable that can be used with both Sonar and Lidar.  At the moment I have chosen to use centimeter level accuracy to ALT_AGL as it is primarily intended to be used closed to the ground for landing purposes. The maximum value of ALT_AGL, as coded,  will be around 30 meters when used with Lidar.

phollands:
MatrixPilot Travis compile checks seems to have disappeared. If anyone knows what has happened, I would be glad to hear from you.

The newly refactored Lidar / Sonar driver code will need flight testing, which I will intend to complete over the next few days. Once I have Lidar flight data, I expect the driver may be changed  to give better confidence in height above ground given better understanding of noise and accuracy.

phollands:
I tested the above commits using Travis in my own account ([build log is here](https://travis-ci.com/phollands/MatrixPilot/builds/123601506)). Passed. (I had to fix the Ubuntu build system in Travis to pull in a developer apt package for the X-Plane plugin in order for the compiler to find GL/gl.h .) 

Travis is working under my own github account, but not for the main MatrixPilot account.

phollands:
Flight tested with Maxbotix Sonar.
The following Logo flight instructions will cut the engine when the plane within 50 centimeters of the ground. (Tested).
```const struct logoInstructionDef instructions[] = {
SET_ALT(10)
SET_INTERRUPT(LAND)
FD(40) 
END

TO (LAND)			          // Interrupt Routine
	IF_LT(ALT_AGL,50)	          // If Altitude Above Ground Level less than 50cm:-
		FLAG_ON(F_LAND)	  //  Engine off
		FD(30) 			  // Interrupt ends if Glide more than 30 meters. (Does not Land).
		CLEAR_INTERRUPT
	END							
END 
```

magicrub:
does Travis have access to this repo? If it works for your repo, but not upstream, then it's likely just a rights issue. Only admins can add or remove access so the list of people able to do that is very small

phollands:
Thanks Tom. Robert Dickenson has fixed up the repo, and MatrixPilot is now re-connected to Travis, and has passed the tests  (As shown by latest commit of a README file on master).

phollands:
I am going to close this Pull Request for now as I after studying the Lidar data from flights in more detail, I think the driver is going to need to be re-worked to effectively work with noise, and to generate a reliable altitude above ground level.

PR #182, opened 2020-01-19
House keeping

Inspirati:
Preliminary commit to reduce the diff on my working branches. Hoping to commit my work on automated JSBSim software in the loop testing sometime in the near future.


Config/Cessna/options.h
```diff
@@ -530,7 +530,7 @@
 // Parameters below are used in the computation of angle of attack and pitch trim.
 // ( INVERTED_NEUTRAL_PITCH is no longer used and should not be used.)
 // If these parameters are not defined, angle of attack and pitch trim will be set to zero.
-// REFERENCE_SPEED                      The nominal speed in meters per second at which the parameters are defined.
+// REFERENCE_SPEED                      The nominal speed in meters per second at which the parameters are defined. (This is only used to predict the stall speed, being half)
```
phollands: Reference Speed is used by helical turns when predicting expected Angle of Attack, and Expected requried Trim for a given airspeed. Reference Speed is used to created a linear interpolation of prediction for AoA and Trim. Helical Turn will disengage it's algorithms when the air speed is less than half of REFERENCE_SPEEED. T

Config/Cessna/options.h
```diff
@@ -530,7 +530,7 @@
 // Parameters below are used in the computation of angle of attack and pitch trim.
 // ( INVERTED_NEUTRAL_PITCH is no longer used and should not be used.)
 // If these parameters are not defined, angle of attack and pitch trim will be set to zero.
-// REFERENCE_SPEED                      The nominal speed in meters per second at which the parameters are defined.
+// REFERENCE_SPEED                      The nominal speed in meters per second at which the parameters are defined. (This is only used to predict the stall speed, being half)
 // ANGLE_OF_ATTACK_NORMAL               Angle of attack in degrees in the body frame for normal straight and level flight at cruise speed.
```
phollands: Cruise speed should read Reference Speed. (and in the next line of comment as well.)

phollands:
Hi Rob, 

I am looking forward to hearing about your Simulator work.

There are a number of points about this pull Request that I would like to discuss or have changed.

- The inference of the title and comments, is that you are trying to get MatrixPilot into a state that allows you to do easier difference analysis on your existing code changes. However, the normal route for commits, is to get the latest copy of Master, and to make your commits additive to the exsiting history of commits.  In this way, history of commits is kept as a linear time ordered set of commits which is easy to understand. Can you "rebase" your commits  onto the existing Master ?

- You have multiple commits to the README.md file in the pull request. For a Pull Request that should be"squashed" (using "interactive rebase" into one commit.  I would also like to understand what the README.md commits are trying to achieve. They change the banner on the home page for the status of the Travis build of MatrixPilot. However that status seems to be working already.

- I note that the your commits have two different identities for Robert Dickenson. I would prefer that you use just one identity.

- REFERENCE_SPEED: Thanks for your most recent commit. After your first change, I took at look at the code in more detail. The code is rather obfuscated in that REFERENCE_SPEED  is halved and becomes stallSpeed. and then stallSpeed is used  as a reference speed (athough it is half), in the [subroutine relativeWindLoading()](https://github.com/phollands/MatrixPilot/blob/master/MatrixPilot/helicalTurnCntrl.c#L164)

It would be nice to chat some time, or feel free to email me one on one. 

Best wishes, Pete

Inspirati:
Work to be posted in new pull request.

PR #186, opened 2020-04-16
Add support for the ICM-20689, for the UDB6mini

benjie-git:


elgarbe:
Is there an udb6?????

benjie-git:
Yes there's a UDB6mini.  I'm in production now.  It's the same as the
UDB5mini, but with an ICM-20689 sensor in place of the deprecated MPU6000.
It has much lower gyro drift, which I find very exciting!  I'll be posting
more info soon...

phollands:
I note that the user will need to configure board type as UDB5, and the code will then automatically work for the UDB6.

chiloschista:
Would then UDB5 flashed still correctly?

⁣Inviato da TypeApp ​

Il giorno 17 Apr 2020, 11:12, alle ore 11:12, Pete Hollands <notifications@github.com> ha scritto:

phollands:
Ric, UDB5 will flash correctly. The code supports both UDB5 and UDB6.

PR #189, opened 2021-04-01
MatrixPilot master branch : minor revisions to pwm computations to in…

wjpremerlani:
…crease range and/or resolution by 1 bit


wjpremerlani:
Two minor improvements to pwm computations:
1. Changed a pwm argument to a subroutine from int16_t to uint16_t. This increases range by one bit.
2. Changed scaling computation to avoid losing one bit in some situations.
These changes have no effect in most cases. However, if the pwm routines are used to measure pulses coming from other sources, such as SONAR or LIDAR, then the changes are useful.

wjpremerlani:
Failed travis build during SILSIM build because of change in data type of pwm variables, so I am closing this pull request.

PR #190, opened 2021-04-02
Fix radio in

wjpremerlani:
Two minor improvements to pwm computations:

Changed a pwm argument to a subroutine from int16_t to uint16_t. This increases range by one bit.
Changed scaling computation to avoid losing one bit in some situations.
These changes have no effect in most cases. However, if the pwm routines are used to measure pulses coming from other sources, such as SONAR or LIDAR, then the changes are useful.


wjpremerlani:
I have run into a bug in my multicopter branch. I am not sure if it has something to do with the subject of this pull request, or if it is some other issue. Mixing signed and unsigned arithmetic is tricky. Anyway, until I resolve the issue that I am having with multicopter, lets hold off merging this pull request.

wjpremerlani:
Team, disregard my last comment. The issue with my multicopter was not the subject of this pull request: I had the connections to two of the motors on my quadcopter swapped.

